PR #5, opened 2025-02-18
Use CircularBuffer to track proven publications

ernestognw:


src/lib/RingBuffer.sol
```diff
@@ -0,0 +1,78 @@
...
+        /// @dev Maps checkpoints to an always increasing index to ensure a checkpoint belongs
+        /// to the current ring (e.g. proving between 3 and 5 will leave indexes 4 and 5 untouched)
+        uint256 index;
+        /// @dev May be anything that describes the rollup state at a given data feed publication (e.g. a state root)
+        bytes32 value;
+    }
+
+    struct CheckpointBuffer {
+        /// @dev Represents the latest overall index (i.e. the index if the buffer didn't wrap around)
+        uint256 _lastIndex;
```
dantaik, line 21: shall we remove leading "_"?

src/protocol/Inbox.sol
```diff
@@ -24,36 +26,66 @@ contract Inbox {
...
     /// @param genesis the checkpoint describing the initial state of the rollup
     /// @param dataFeed the input data source that updates the state of this rollup
     /// @param verifier a contract that can verify the validity of a transition from one checkpoint to another
-    constructor(bytes32 genesis, address dataFeed, address verifier) {
+    constructor(uint256 bufferSize, bytes32 genesis, address dataFeed, address verifier) {
         // set the genesis checkpoint of the rollup - genesis is trusted to be correct
         require(genesis != 0, "genesis checkpoint cannot be 0");
-        checkpoints[0] = genesis;
+        _checkpoints.setup(bufferSize);
+        _checkpoints.setAt(0, genesis).setLastIndex(1);
```
dantaik, line 40: should we use `setLastIndex(0)` instead? The last index is index 0 which is the genesis.

src/lib/RingBuffer.sol
ernestognw: I use to prefix with `_` when the developers are not supposed to access the value directly (as in saying that it's private). No strong opinion here so I'll update.

src/lib/RingBuffer.sol
```diff
@@ -0,0 +1,78 @@
...
+        clear(self);
+        Checkpoint[] storage _checkpoints = self._checkpoints;
+        assembly ("memory-safe") {
+            sstore(_checkpoints.slot, size)
+        }
+    }
+
+    /// Getters
+
+    function clear(CheckpointBuffer storage self) internal {
```
dantaik, line 37: Maybe we can simply remove `clear()` and use `setLastIndex` directly instead.

src/lib/RingBuffer.sol
```diff
@@ -0,0 +1,78 @@
...
+        self._lastIndex = 0;
+    }
+
+    function length(CheckpointBuffer storage self) internal view returns (uint256) {
```
dantaik, line 41: I wish this length can be a immutable :D

src/protocol/Inbox.sol
```diff
@@ -24,36 +26,66 @@ contract Inbox {
...
-        require(checkpoint != 0, "Checkpoint cannot be 0");
-        require(end > lastProvenIdx, "Publication already proven");
-        bytes32 base = checkpoints[start];
-        // this also ensures start <= lastProvenIdx
-        require(base != 0, "Unknown base checkpoint");
+        RingBuffer.Checkpoint storage startCheckpoint = _checkpoints.at(start);
+
+        // Checks
+        uint256 lastIndex = _checkpoints.lastIndex();
+        require(start < lastIndex && startCheckpoint.index == start, NotProven(start));
```
dantaik, line 75:
```suggestion
        require(start <= lastIndex && startCheckpoint.index == start, NotProven(start));
```

src/protocol/Inbox.sol
```diff
@@ -24,36 +26,66 @@ contract Inbox {
...
+        require(end >= lastIndex, AlreadyProven(end));
```
dantaik, line 76:
```suggestion
        require(end > lastIndex, AlreadyProven(end));
```

src/protocol/Inbox.sol
```diff
@@ -24,36 +26,66 @@ contract Inbox {
...
+        _checkpoints.setAt(0, genesis).setLastIndex(1);
         _dataFeed = IDataFeed(dataFeed);
         _verifier = IVerifier(verifier);
     }
 
+    /// @notice Returns the checkpoint at the given publication index
+    /// @param at The index to query
+    /// @return The checkpoint at the given index (wrapped around buffer length)
+    function getCheckpoint(uint256 at) public view returns (bytes32) {
+        return _checkpoints.at(at).value;
```
dantaik, line 49: I still tahink we should check `_checkpoints.at(at).index == at`

PR #16, opened 2025-02-20
[DRAFT] allow each rollup to define when to sync direct publications

ggonzalez94:
This is a proposal for how we can allow each rollup(via its provider, which we should rename to hooks or similar) what direct publication they want to reference. This avoids enforcing "real time" CR inclusion for rollups, and gives them the flexibility to define rules about when a direct publication is linked to an actual publication. 
This will usually be time delays, similar to a delayed inbox, but in theory a rollup can define more complex rules if they want.

This is how a user would force a tx:
1. They(or someone on their behalf) calls `directPublish` with bundles for one or multiple rollups
2. They(or someone on their behalf) calls `setLastDirectPublicationHash` on the provider to update the last direct publication that the rollup "sees". Here each provider can enforce a delay, or whatever rule they want.
3. The direct publication is linked to the next publication when the metadata is retrieved.

I'm not sure if this is the best approach, or if it is just better to keep separate queues per rollup. But this seemed simple and elegant enough that it made sense to me :)

Once we agree with an approach I can go ahead and rename the provider to Hooks or something similar.

### DIFF
--- a/src/protocol/DataFeed.sol
+++ b/src/protocol/DataFeed.sol
@@ -36,7 +36,6 @@ contract DataFeed is IDataFeed {
         require(numBlobs > 0, "no data to publish");
 
         uint256 id = publicationHashes.length;
-        uint256 directId = directPublicationHashes.length;
         Publication memory publication = Publication({
             id: id,
             prevHash: publicationHashes[id - 1],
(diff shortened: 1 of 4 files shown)


ggonzalez94:
Closing this since we have a new design for delayed publications

PR #17, opened 2025-02-20
Calldata Support

nikeshnazareth:
Include calldata in the Publication.

I also replaced the `blobHashes` field with a single data hash that covers all blobs and calldata.

Also minor cleanup: use inheritdoc for the NatSpec comments

### DIFF
--- a/src/protocol/DataFeed.sol
+++ b/src/protocol/DataFeed.sol
@@ -36,22 +36,21 @@ contract DataFeed is IDataFeed {
         uint256 nQueries = queries.length;
         uint256 id = publicationHashes.length;
 
-        bytes32[] memory blobHashes = new bytes32[](numBlobs);
-        for (uint256 i; i < numBlobs; ++i) {
-            blobHashes[i] = blobhash(i);
-        }
-        bytes32 dataHash = keccak256(abi.encode(data, blobHashes));
-
         Publication memory publication = Publication({
             id: id,
             prevHash: publicationHashes[id - 1],
             publisher: msg.sender,
             timestamp: block.timestamp,
             blockNumber: block.number,
-            dataHash: dataHash,
+            blobHashes: new bytes32[](numBlobs),
+            data: data,
             queries: queries,
             metadata: new bytes[](nQueries)
         });
+
+        for (uint256 i; i < numBlobs; ++i) {
+            publication.blobHashes[i] = blobhash(i);
+        }
 
         uint256 totalValue;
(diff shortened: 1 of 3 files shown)


src/protocol/DataFeed.sol
```diff
@@ -27,33 +27,34 @@ contract DataFeed is IDataFeed {
...
-    /// @dev append a hash representing all blobs and L1 metadata to `publicationHashes`.
-    /// The number of blobs is not validated. Additional blobs are ignored. Empty blobs have a hash of zero.
-    function publish(uint256 numBlobs, MetadataQuery[] calldata queries) external payable onlyStandaloneTx {
+    /// @inheritdoc IDataFeed
+    function publish(uint256 numBlobs, bytes calldata data, MetadataQuery[] calldata queries)
+        external
+        payable
+        onlyStandaloneTx
+    {
         require(numBlobs > 0, "no data to publish");
```
ggonzalez94, line 36: Should we remove the requirement that you have to include blobs? You might want to only include calldata

src/protocol/DataFeed.sol
```diff
@@ -27,33 +27,34 @@ contract DataFeed is IDataFeed {
...
+        bytes32 dataHash = keccak256(abi.encode(data, blobHashes));
+
         Publication memory publication = Publication({
             id: id,
             prevHash: publicationHashes[id - 1],
             publisher: msg.sender,
             timestamp: block.timestamp,
             blockNumber: block.number,
-            blobHashes: new bytes32[](numBlobs),
+            dataHash: dataHash,
```
dantaik, line 53: do we want to emit the `data` as part of the event? If not, and when `publish` is called in wrapping transaction, it's really hard to extract the `data` from the inner transaction -- even if we use use geth's tracing API.

ggonzalez94 (commented):
Looks good. Just one small suggestion

src/protocol/DataFeed.sol
dantaik: If later we remove the following:

```
bytes32[] memory blobHashes = new bytes32[](numBlobs);
        for (uint256 i; i < numBlobs; ++i) {
            blobHashes[i] = blobhash(i);
        }
```

We can remove the `uint256 numBlobs` input in the `publish` function, then maybe we can simply emit the `data` as part of the publication, not its hash.

ggonzalez94 (approved):
LGTM

dantaik (approved):

PR #18, opened 2025-02-20
Rename metadata provider to hook provider

ggonzalez94:
Naively rename everything related to metadata to hooks and make the minimal changes required in the `DataFeed` to use the hooks.
Since we will refactor many changes in the DataFeed and how we interpret and handle publications I didn't spend much time trying to accommodate for that. I reused the signature of the `before_publish` hook for the `after_publish` hook, but we may want to change that once we have a clearer design for how we handle the forced txs.


src/protocol/IHookProvider.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+interface IHookProvider {
```
dantaik, line 4: how about just `IHook`? `IHookProvider` sounds like it will provide one or more hooks.

ggonzalez94: What about `IPublicationHook`? For some reason I feel `IHook` is too generic and too short, but not a strong opinion 😄

src/protocol/IHookProvider.sol
```diff
@@ -0,0 +1,16 @@
...
+    /// @notice Hook called before a publication
+    /// @param publisher The address of the publisher
+    /// @param input Arbitrary input to the hook
+    /// @return output Arbitrary output from the hook. Commonly used to return the metadata for the publication.
+    function before_publish(address publisher, bytes memory input) external payable returns (bytes memory output);
```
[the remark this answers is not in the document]
ggonzalez94: good catch

ggonzalez94:
I think this should be ready for final review and merge now

dantaik (approved):

src/protocol/DataFeed.sol
```diff
@@ -38,6 +38,17 @@ contract DataFeed is IDataFeed {
         require(numBlobs > 0, "no data to publish");
 
         uint256 nQueries = queries.length;
+
+        uint256 totalValue;
+        bytes[] memory metadata = new bytes[](nQueries);
+        for (uint256 i; i < nQueries; ++i) {
+            metadata[i] = IPublicationHook(queries[i].provider).beforePublish{value: queries[i].value}(
+                msg.sender, queries[i].input
```
dantaik, line 46: shall we split the input into preInput and postInput?

dantaik: I think I may slight prefer an alternative implementation: adding a flag to the Query to indicate whether this query should run before or after the publication. So we have one not two hook functions in the hook interface:

```solidity

interface IPublicationHook {
function someName(address publisher, bool isPriorToPublication, bytes memory input) external payable returns (bytes memory output);
}

```

src/protocol/DataFeed.sol
```diff
@@ -38,6 +38,17 @@ contract DataFeed is IDataFeed {
...
+            );
+            totalValue += queries[i].value;
```
dantaik, line 48: beforePublish and postPublish may use different value...

nikeshnazareth: can we just avoid sending ETH in the post-publish case? Any ETH it needs can be passed along with the pre-publish hook, and the hook contract can partition the value if necessary.

nikeshnazareth: I agree that having different pre-post hook queries would be cleaner.

src/protocol/DataFeed.sol
```diff
@@ -14,8 +14,22 @@ contract DataFeed is IDataFeed {
     }
 
     /// @inheritdoc IDataFeed
-    function publish(uint256 numBlobs, bytes calldata data, MetadataQuery[] calldata queries) external payable {
-        uint256 nQueries = queries.length;
+    function publish(
+        uint256 numBlobs,
+        bytes calldata data,
+        HookQuery[] calldata metadataQueries,
```
dantaik, line 20: remove all occurances of "metadata"?

src/protocol/IDataFeed.sol
```diff
@@ -16,7 +16,8 @@ interface IDataFeed {
         uint256 blockNumber;
         bytes32[] blobHashes;
         bytes data;
-        MetadataQuery[] queries;
+        HookQuery[] metadataQueries;
```
dantaik, line 19: beforeHooks

src/protocol/DataFeed.sol
nikeshnazareth: isn't it actually returning the L1 metadata?

nikeshnazareth (approved):

src/protocol/IDataFeed.sol
```diff
@@ -16,8 +16,9 @@ interface IDataFeed {
         uint256 blockNumber;
         bytes32[] blobHashes;
         bytes data;
-        MetadataQuery[] queries;
-        bytes[] metadata;
+        HookQuery[] preHooks;
+        HookQuery[] postHooks;
```
ggonzalez94, line 20: nit:
```suggestion
        HookQuery[] preHooksQuery;
        HookQuery[] postHooksQuery;
```

src/protocol/DataFeed.sol
```diff
@@ -14,8 +14,13 @@ contract DataFeed is IDataFeed {
     }
 
     /// @inheritdoc IDataFeed
-    function publish(uint256 numBlobs, bytes calldata data, MetadataQuery[] calldata queries) external payable {
-        uint256 nQueries = queries.length;
+    function publish(
+        uint256 numBlobs,
+        bytes calldata data,
+        HookQuery[] calldata preHooks,
+        HookQuery[] calldata postHooks
```
ggonzalez94, line 21: nit:
```suggestion
        HookQuery[] calldata preHooksQuery,
        HookQuery[] calldata postHooksQuery
```

ggonzalez94 (commented):
Looks ready to go. Last few nits(we can always change those when we clean variable names, etc. if you prefer)

ernestognw (approved):

PR #19, opened 2025-02-20
Remove onlyStandaloneTx requirement

nikeshnazareth:
We intend to let the hooks invoke the publish function.
We will need to introduce a different mechanism to prevent spam.


ggonzalez94 (approved):
LGTM

dantaik (approved):

PR #22, opened 2025-02-20
Add ERC7786 Gateways

LeoPatOZ:
- The idea is to add ERC7786 gateways for source and destination


ggonzalez94:
Closing this since I don't think we have plans to include this natively in the protocol

PR #24, opened 2025-02-20
Re-implement TaikoHookProvider

dantaik:


src/protocol/taiko_alethia/TaikoPublicationHook.sol
```diff
@@ -1,42 +1,89 @@
...
+        require(maxAnchorBlockOffset + _input.anchorBlockId >= block.number, "anchorBlockId too old");
+        require(_input.numBlobs > 0, "no blobs used");
+
+        bytes32 anchorBlockhash = blockhash(_input.anchorBlockId);
         require(anchorBlockhash != 0, "blockhash not found");
-        return abi.encode(anchorBlockhash);
+
+        bytes32[] memory blobs = new bytes32[](_input.numBlobs);
+        for (uint8 i; i < _input.numBlobs; i++) {
+            blobs[i] = blockhash(_input.anchorBlockId + i);
```
nikeshnazareth, line 60: I don't think I understand this line. Should it be `blobhash(i)`?

src/protocol/taiko_alethia/ITaikoData.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+interface ITaikoData {
+    /// @dev Represents a data source for a proposal, containing multiple blobs.
+    /// These blobs are concatenated into a single blob or string, which then is decoded into a specific format as
+    /// defined by the node or client.
+    struct DataSource {
+        bytes32[] blobs;
```
ernestognw, line 9: I find confusing to call these `blobs`. Shoulnd't they be `blobhashes`?

src/protocol/taiko_alethia/ITaikoData.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+interface ITaikoData {
+    /// @dev Represents a data source for a proposal, containing multiple blobs.
+    /// These blobs are concatenated into a single blob or string, which then is decoded into a specific format as
+    /// defined by the node or client.
```
ernestognw, line 7:
```suggestion
    /// @dev Represents a data source for a proposal, containing multiple blobhashes
    /// These hashes are concatenated into a single blob or string, which then is decoded into a specific format as
    /// defined by the node or client.
```

nikeshnazareth:
Obsolete by #29

PR #29, opened 2025-02-21
Remove data feed hooks

nikeshnazareth:
The new design restores the DataFeed to a simply list of publications.

The logic for constructing publications, including the delayed inclusion mechanism
is moved to the TaikoInbox.

Rename the Inbox contract to Checkpoint to avoid confusion, since it does not receive
user messages


src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -0,0 +1,59 @@
...
+
+        bytes[] memory attributes = new bytes[](1);
+        attributes[0] = abi.encode(proposal);
+        datafeed.publish(attributes);
+
+        // reuse proposal anchorBlockHash
+        proposal.proposalDataList = delayedInclusionStore.processDelayedInclusionByDeadline(block.timestamp);
+        if (proposal.proposalDataList.length > 0) {
+            attributes[0] = abi.encode(proposal);
+            datafeed.publish(attributes);
```
LeoPatOZ, line 56: Perhaps we can merge the forced inclusion list to the data source and only call publish once

src/protocol/Checkpoint.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.28;
 import {IDataFeed} from "./IDataFeed.sol";
 import {IVerifier} from "./IVerifier.sol";
 
-contract Inbox {
+contract Checkpoint {
```
ernestognw, line 7: I would reconsider renaming everything inside that we're calling a `checkpoint`. Given the contract itself is called `checkpoint`, everything else may be deemed redundant

src/protocol/IDataFeed.sol
```diff
@@ -2,44 +2,25 @@
...
-        uint256 value;
-    }
-
-    struct Publication {
+    struct PublicationHeader {
         uint256 id;
         bytes32 prevHash;
         address publisher;
         uint256 timestamp;
         uint256 blockNumber;
```
ernestognw, line 10: We can pack publisher, blockNumber and timestamp within a single slot (i.e. (uint160, uint48, uint48)). Not to do it now as the design may change again but to consider later.

src/protocol/DataFeed.sol
```diff
@@ -14,53 +13,24 @@ contract DataFeed is IDataFeed {
...
-            preHookQueries: preHookQueries,
-            postHookQueries: postHookQueries,
-            auxData: auxData
+            blockNumber: block.number
         });
 
-        for (uint256 i; i < numBlobs; ++i) {
-            publication.blobHashes[i] = blobhash(i);
+        bytes32[] memory attributeHashes = new bytes32[](attributes.length);
+        for (uint256 i; i < attributes.length; ++i) {
```
ernestognw: May be cool to cache the length

```suggestion
        uint256 attributesLength = attributes.length;
        bytes32[] memory attributeHashes = new bytes32[](attributesLength);
        for (uint256 i; i < attributesLength; ++i) {
```

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -0,0 +1,59 @@
...
+        address _delayedInclusionStore,
+        uint256 _maxAnchorBlockIdOffset
+    ) {
+        datafeed = IDataFeed(_datafeed);
+        lookahead = ILookahead(_lookahead);
+        delayedInclusionStore = IDelayedInclusionStore(_delayedInclusionStore);
+        maxAnchorBlockIdOffset = _maxAnchorBlockIdOffset;
+    }
+
+    function publish(uint256 nBlobs, bytes calldata data, uint64 anchorBlockId) external {
```
LeoPatOZ, line 29: Can we change this function to `propose` given we are dealing with a proposal at this stage

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -0,0 +1,70 @@
...
+import {ITaikoData} from "./ITaikoData.sol";
+
+contract TaikoInbox {
+    IDataFeed public immutable datafeed;
+    ILookahead public immutable lookahead;
+    IDelayedInclusionStore public immutable delayedInclusionStore;
+
+    uint256 public immutable maxAnchorBlockIdOffset;
+
+    uint256 public prevPublicationId;
```
ggonzalez94, line 19: Should we consider renaming this to `publicationId` or `currentPublicationId`? Because in reality it is the current publication id at the moment that you call `publish` and only then it gets overwritten?

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -0,0 +1,70 @@
...
+        maxAnchorBlockIdOffset = _maxAnchorBlockIdOffset;
+    }
+
+    function publish(uint256 nBlobs, uint64 anchorBlockId) external {
+        if (address(lookahead) != address(0)) {
+            require(lookahead.isCurrentPreconfer(msg.sender), "not current preconfer");
+        }
+
+        bytes[] memory attributes = new bytes[](3);
+        uint256 _prevPublicationId = prevPublicationId;
```
ggonzalez94, line 41: Same as the last comment, maybe we just name this variable _publicationId?

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -0,0 +1,70 @@
...
+            require(blobHashes[i] != 0, "data unavailable");
+        }
+        attributes[2] = abi.encode(blobHashes);
+        _prevPublicationId = datafeed.publish(attributes).id;
+
+        // Publish each inclusion as a publication
+        ITaikoData.DataSource[] memory dataSources =
+            delayedInclusionStore.processDelayedInclusionByDeadline(block.timestamp);
+
+        uint256 nDataSources = dataSources.length;
```
ggonzalez94: I don't think we need the extra variable

src/protocol/taiko_alethia/TaikoInbox.sol
nikeshnazareth: we decided against this. We should create a design doc shortly to explain the rationale.

ggonzalez94 (commented):
Looks great overall, let a few suggestions

src/protocol/taiko_alethia/TaikoInbox.sol
dantaik: When a new publication is going to be published, I consider this new publication to be the "current" one, then we need to link this current on to the previous (the last) publication of the same kind (from the same msg.sender), so that's why it's called prePublicationId, we can also call it "lastPublicationId"

dantaik (approved):

ernestognw (approved):
LGTM with a small nit: we're still calling `pubHash` what I think is now the `headerHash`. There are some implication of that renaming that I don't think it's worth going through here.

PR #30, opened 2025-02-21
Introduce BlobRegistry

dantaik:
## About

Many dApps rely on blobs for data availability (DA) and require off-chain software to locate and verify data within them. `BlobSourceProvider` allows dApps to generate `BlobSource` objects, emit events containing them, and store only their hashes on-chain for later reuse. In most cases, `BlobSource` objects themselves should not be stored on-chain.

@ggonzalez94 @nikeshnazareth @ernestognw  I could use some help refining the contract and function names.


ggonzalez94:
@dantaik did some updates:
- changed BlobSourceProvider -> BlobRegistry
- added natspec
- use uint256 in the function signatures instead of uint16 since there's no gas savings

src/blobs/BlobRegistry.sol
```diff
@@ -0,0 +1,55 @@
...
+
+import "./IBlobRegistry.sol";
+
+/// @title BlobRegistry
+contract BlobRegistry is IBlobRegistry {
+    /// @dev A mapping of the hash of a blob source to the timestamp when it was saved
+    mapping(bytes32 sourceHash => uint256 timestamp) private _savedHashes;
+
+    /// @inheritdoc IBlobRegistry
+    function getAndSaveHash(uint256[] calldata blobIdxs)
```
ernestognw, line 12: If we assume there are not going to be more than 255 blobs, then it's possible to express this array as a `bytes32 blobIdxs` bitmap so that we don't allocate calldata for the indexes. For example, instead of `[0,1,2]` we can provide `0x111000...`

ernestognw (commented):
Added naming suggestions, may be confusing but the general idea is that the contract should not imply that it's retrieving the "blob data" since that's not possible. Similarly, we're referring to the "source hash" as just "hash"

src/blobs/BlobRegistry.sol
dantaik: Previously I did have another function `function getAndSaveHash(bytes32 encodedBlobIndices)` but I deleted as I don't know if we need it here or a calling dapp shall encode the indices in its function.

src/blobs/IBlobRegistry.sol
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /// @notice Retrieves blob data for given blob indices and saves their hash
+    /// @param blobIdxs Array of blob indices to retrieve
+    /// @return blobSource The retrieved blob data including block number and blob hashes
+    /// @return hash The keccak256 hash of the encoded blob source
+    /// @dev Should revert if any blob index is invalid or if no blobs are provided
+    function getAndSaveHash(uint256[] calldata blobIdxs)
+        external
+        returns (BlobSource memory blobSource, bytes32 hash);
```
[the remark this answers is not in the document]
ggonzalez94: why sourceHash? I think even just hash is clearer? I'm open to blobhash perhaps?

src/blobs/BlobRegistry.sol
ernestognw: I understand the dapp should encode the indices on their side, so I agree it's not ideal. Current Implementation is good.

ernestognw (approved):

src/blobs/BlobRegistry.sol
dantaik: See https://github.com/OpenZeppelin/minimal-rollup/pull/33

PR #31, opened 2025-02-22
Suggest a few more changes 

dantaik:


src/protocol/Checkpoint.sol
```diff
@@ -40,19 +41,19 @@ contract Checkpoint {
     /// @param newCheckpoint The proposed new checkpoint value to transition to
     /// @param proof Arbitrary data passed to the `_verifier` contract to confirm the transition validity.
     function proveTransition(uint256 end, bytes32 newCheckpoint, bytes calldata proof) external {
-        require(checkpoint != 0, "Checkpoint cannot be 0");
-        require(end > lastProvenIdx, "Publication already proven");
+        require(newCheckpoint != 0, "Checkpoint cannot be 0");
```
dantaik, line 44: a small bug fix here

src/protocol/DataFeed.sol
```diff
@@ -13,21 +13,22 @@ contract DataFeed is IDataFeed {
...
 
-        bytes32[] memory attributeHashes = new bytes32[](attributes.length);
-        for (uint256 i; i < attributes.length; ++i) {
+        uint256 nAttributes = attributes.length;
+        bytes32[] memory attributeHashes = new bytes32[](nAttributes);
+        for (uint256 i; i < nAttributes; ++i) {
             attributeHashes[i] = keccak256(attributes[i]);
         }
-        bytes32 pubHash = keccak256(abi.encode(header, attributeHashes));
+        bytes32 pubHash = keccak256(abi.encode(header, keccak256(abi.encode(attributeHashes))));
```
dantaik, line 31: hash the attributesHashes first.

ernestognw: Just to confirm, is the reason of this change to decouple the attributeHashes to its hash so that someone may decide not to store the list of attributeHashes?

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -26,34 +28,44 @@ contract TaikoInbox {
         maxAnchorBlockIdOffset = _maxAnchorBlockIdOffset;
     }
 
-    function publish(uint256 nBlobs, bytes calldata data, uint64 anchorBlockId) external {
+    function publish(uint256 nBlobs, uint64 anchorBlockId) external {
         if (address(lookahead) != address(0)) {
             require(lookahead.isCurrentPreconfer(msg.sender), "not current preconfer");
         }
```
ernestognw, line 34: Isn't this redundant? The lookahead should return false for the `address(0)`

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -26,34 +28,44 @@ contract TaikoInbox {
         maxAnchorBlockIdOffset = _maxAnchorBlockIdOffset;
     }
 
-    function publish(uint256 nBlobs, bytes calldata data, uint64 anchorBlockId) external {
+    function publish(uint256 nBlobs, uint64 anchorBlockId) external {
```
ernestognw, line 31: This is not an issue using our canonical data feed. However, if a user comes with a custom implementation, there might be side effects they must be aware of

```suggestion
    /// WARNING: Make sure `datafeed` is trusted given it can reenter when calling `publish` while the state
    /// is inconsistent (i.e. `prevPublicationId` has not been set)
    function publish(uint256 nBlobs, uint64 anchorBlockId) external {
```

src/protocol/taiko_alethia/TaikoInbox.sol
nikeshnazareth: I think that's true for a low-level call. The solc compiler will add a check to ensure `lookahead` is a valid contract and revert if not.

src/protocol/taiko_alethia/TaikoInbox.sol
ernestognw: Same for the `delayedInclusionStore`

PR #34, opened 2025-02-22
Use IBlobRefRegistry in TaikoInbox

dantaik:


src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -19,11 +21,13 @@ contract TaikoInbox {
     constructor(
         address _datafeed,
         address _lookahead,
+        address _blobRefRegister,
```
ernestognw, line 24: Nit, but, any reason to name it "register"?

```suggestion
        address _blobRefRegistry,
```

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -19,11 +21,13 @@ contract TaikoInbox {
...
         address _delayedInclusionStore,
         uint256 _maxAnchorBlockIdOffset
     ) {
         datafeed = IDataFeed(_datafeed);
         lookahead = ILookahead(_lookahead);
+        blobRefRegister = IBlobRefRegistry(_blobRefRegister);
```
ernestognw, line 30:
```suggestion
        blobRefRegistry = IBlobRefRegistry(_blobRefRegistry);
```

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -1,15 +1,17 @@
...
 import {IDataFeed} from "../IDataFeed.sol";
 
 import {IDelayedInclusionStore} from "./IDelayedInclusionStore.sol";
 import {ILookahead} from "./ILookahead.sol";
-import {ITaikoData} from "./ITaikoData.sol";
 
 contract TaikoInbox {
     IDataFeed public immutable datafeed;
     ILookahead public immutable lookahead;
+    IBlobRefRegistry public immutable blobRefRegister;
```
ernestognw, line 14:
```suggestion
    IBlobRefRegistry public immutable blobRefRegistry;
```

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -45,26 +49,27 @@ contract TaikoInbox {
...
+        IBlobRefRegistry.BlobRef[] memory blobRefs =
             delayedInclusionStore.processDelayedInclusionByDeadline(block.timestamp);
 
-        uint256 nDataSources = dataSources.length;
-        for (uint256 i; i < nDataSources; ++i) {
+        uint256 nBlobRefs = blobRefs.length;
+        for (uint256 i; i < nBlobRefs; ++i) {
             attributes[1] = abi.encode(_prevPublicationId);
-            attributes[2] = abi.encode(dataSources[i].blobHashes);
+            attributes[2] = abi.encode(blobRefs[i]);
```
ernestognw, line 62: Why do we need to encode all the blob ref when the next line (i.e. `datafeed.publish`) will store the block number?

ggonzalez94 (approved):
LGTM

src/protocol/taiko_alethia/TaikoInbox.sol
ggonzalez94: In the case of a delayed inbox we do need it, so it makes sense to have the same struct

ernestognw: @nikeshnazareth explained to me that there may be publications with references to other blocks for the delayed inbox. That makes sense. Solving

PR #36, opened 2025-02-22
Add getNextPublicationId() to DataFeed

dantaik:


src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -49,7 +49,7 @@ contract TaikoInbox {
         require(anchorBlockId >= block.number - maxAnchorBlockIdOffset, "anchorBlockId is too old");
         bytes32 anchorBlockhash = blockhash(anchorBlockId);
         require(anchorBlockhash != 0, "blockhash not found");
-        attributes[ANCHOR_TX] = abi.encode(anchorBlockId, anchorBlockhash);
+        attributes[ANCHOR_TX] = abi.encode(datafeed.getNextPublicationId(), anchorBlockhash);
```
nikeshnazareth, line 52: If we make this change to the semantics, should we also update the next publication Id in the delayed inclusion publications below?

dantaik: Oh, I think we can avoid adding datafeed.getNextPublicationId() there as the publication ID is already part of the publication.

ernestognw (approved):

nikeshnazareth (approved):

PR #39, opened 2025-02-22
Some TaikoInbox improvements

dantaik:


src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -42,32 +48,36 @@ contract TaikoInbox {
...
         // Publish each delayed inclusion as a separate publication
-        IBlobRefRegistry.BlobRef[] memory blobRefs =
-            delayedInclusionStore.processDelayedInclusionByDeadline(block.timestamp);
+        IDelayedInclusionStore.Inclusion[] memory inclusions = delayedInclusionStore.processDueInclusions();
+        uint256 nInclusions = inclusions.length;
+        metadata.isDelayedInclusion = true;
+        for (uint256 i; i < nInclusions; ++i) {
+            attributes[METADATA] = abi.encode(metadata);
+            attributes[LAST_PUBLICATION] = abi.encode(_lastPublicationId);
+            attributes[BLOB_REFERENCE] = abi.encode(inclusions[i]);
```
nikeshnazareth, line 75: Does it matter that this is a different type to the usual `BLOB_REFERENCE`?

A normal `BLOB_REFERENCE` attribute is the abi encoding of a `BlobRef` object. Within this loop, the reference is the abi encoding of an `Inclusion` object (which contains the BlobRef hash, not the full BlobRef).

Perhaps it's fine because the `isDelayedInclusion` flag can be used to recognise how to interpret it.

dantaik: I think it should be fine. It will be safe to assume in the future the forced inclusion data will use a different format/schema than the regular block proposal blobs.

src/blobs/BlobRefRegistry.sol
```diff
@@ -12,6 +12,7 @@ contract BlobRefRegistry is IBlobRefRegistry {
     function registerRef(uint256[] calldata blobIndices) external returns (bytes32 refHash, BlobRef memory ref) {
         ref = _getRef(blobIndices);
         refHash = _registerRefHash(ref);
+        emit Registered(refHash, ref);
```
nikeshnazareth, line 15: I removed this event in PR 38 because it is emitted in `_registerRefHash`. Are you intentionally adding it back (because I misunderstood something) or was this just an artifact of the parallel PRs?

nikeshnazareth (approved):

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -42,32 +48,36 @@ contract TaikoInbox {
...
-        bytes32 anchorBlockhash = blockhash(anchorBlockId);
-        require(anchorBlockhash != 0, "blockhash not found");
-        attributes[ANCHOR_TX] = abi.encode(anchorBlockId, anchorBlockhash);
+        require(anchorBlockId >= block.number - maxAnchorBlockIdOffset, "anchorBlockId too old");
 
-        // Build the attribute to link back to the previous publication Id;
-        attributes[PREV_PUBLICATION] = abi.encode(_lastPublicationId);
+        Metadata memory metadata;
+        metadata.anchorBlockId = anchorBlockId;
+        metadata.anchorBlockHash = blockhash(anchorBlockId);
```
ggonzalez94, line 58: Nit:
I don't think we save any gas by doing field by field and leaving the last one by default, the compiler is smart enough to do this optimization when it is a memory struct. So this is a bit more readable IMO:
```suggestion
        Metadata memory metadata = Metadata({
            anchorBlockId: anchorBlockId,
            anchorBlockHash: blockhash(anchorBlockId),
            isDelayedInclusion: false
        });
```

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -42,32 +48,36 @@ contract TaikoInbox {
...
         attributes[BLOB_REFERENCE] = abi.encode(blobRefRegistry.getRef(_buildBlobIndices(nBlobs)));
+
         _lastPublicationId = publicationFeed.publish(attributes).id;
 
         // Publish each delayed inclusion as a separate publication
-        IBlobRefRegistry.BlobRef[] memory blobRefs =
-            delayedInclusionStore.processDelayedInclusionByDeadline(block.timestamp);
+        IDelayedInclusionStore.Inclusion[] memory inclusions = delayedInclusionStore.processDueInclusions();
+        uint256 nInclusions = inclusions.length;
+        metadata.isDelayedInclusion = true;
```
ggonzalez94, line 71: Just for clarity I would add a comment on top explaining that the metadata is the same for the set of forced inclusions, we just set the flag to true
```suggestion
        // Metadata is the same as the regular publication, so we just set `isDelayedInclusion` to true
        metadata.isDelayedInclusion = true;
```

ggonzalez94 (approved):
Looks good! Just a couple very minor nits

src/blobs/BlobRefRegistry.sol
dantaik: I added it back intentionally as I feel it will be more convenient for apps to subscribe to event from just this contract to get the ref objects.

dantaik: Since this is the only outstanding issue, I'll merge the PR first then we can change it later if you prefer this event should be removed.

PR #41, opened 2025-02-22
Implement Signal Service using Merkle Mountain Ranges (MMR)

dantaik:
@ernestognw Using this PR I tried to front-run you with your idea of using MMR to implement the signal service :D

UPDATE:
The library I'm using is not optimized for our use case, please do not review this PR.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = lib/forge-std
 	url = https://github.com/foundry-rs/forge-std
 
+[submodule "lib/solidity-mmr"]
+	path = lib/solidity-mmr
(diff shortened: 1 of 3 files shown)


dantaik:
1.	Number of Roots:
With ￼ elements, you need to store as many roots as there are 1’s in the binary representation of ￼.
	2.	Roots Updated on Insertion:
When inserting a new element, the number of roots that may need to be updated is equal to the number of trailing 1’s in the binary representation of ￼; in the worst case, this is ￼.

This explains both the storage and update complexity of the Merkle Mountain Range with respect to the number of elements it contains.

dantaik:
MMR-based impl won't work!!

PR #42, opened 2025-02-23
Parallel Verifications

nikeshnazareth:
A proposal to track proven transitions in parallel.

It will need to be extended to track the prover once the proving incentive mechanism is finalised


src/protocol/ICheckpointTracker.sol
```diff
@@ -0,0 +1,26 @@
...
+    /// @param startPublicationId the index of the publication before the transition
+    /// @param startCommitment the checkpoint at the start of the transition
+    /// @param endPublicationId the index of the last publication in the transition
+    /// @param endCommitment the checkpoint after the transition
+    event TransitionProven(
+        uint256 indexed startPublicationId,
+        bytes32 indexed startCommitment,
+        uint256 indexed endPublicationId,
+        bytes32 endCommitment
+    );
```
LeoPatOZ, line 25: Shouldnt we also add:
`proveTransition`
`advanceTo`

ggonzalez94 (commented):
Looks good, just one suggestion and we can also just store the hash of the `provenCheckpoint` instead of the entire checkpoint

src/protocol/CheckpointTracker.sol
```diff
@@ -0,0 +1,85 @@
...
+            publicationFeed.getPublicationHash(start.publicationId),
+            publicationFeed.getPublicationHash(end.publicationId),
+            start.commitment,
+            end.commitment,
+            proof
+        );
+
+        if (start.publicationId == provenCheckpoint.publicationId && start.commitment == provenCheckpoint.commitment) {
+            provenCheckpoint = end;
+            emit CheckpointProven(end.publicationId, end.commitment);
```
dantaik, line 61: Shouldn't we reuse the same CheckpointProven event with the 4 arguments?

nikeshnazareth: Do you mean the `TransitionProven` event? We are emitting `TransitionProven` events whenever a valid transition is recorded, whether or not it eventually connects with the actual state changes. `CheckpointProven` implies that it's part of the correct chain.

src/protocol/CheckpointTracker.sol
```diff
@@ -0,0 +1,81 @@
...
+    function proveTransition(Checkpoint calldata start, Checkpoint calldata end, bytes calldata proof) external {
+        bytes32 startCheckpointHash = keccak256(abi.encode(start));
+        bytes32 endCheckpointHash = keccak256(abi.encode(end));
+
+        require(end.commitment != 0, "Checkpoint commitment cannot be 0");
+        // TODO: once the proving incentive mechanism is in place we should reconsider this requirement because
+        // ideally we would use the proof that creates the longest chain of proven publications.
+        require(transitions[startCheckpointHash] == 0, "Checkpoint already has valid transition");
+        require(start.publicationId < end.publicationId, "Start must be before end");
+        require(end.publicationId < publicationFeed.getNextPublicationId(), "Publication does not exist");
```
dantaik, line 51: instead of this check which requires an `getNextPublicationId` call, maybe we can simply check `getPublicationHash(start.publicationId) !=0  &&getPublicationHash(end.publicationId) !=0`

src/protocol/CheckpointTracker.sol
```diff
@@ -0,0 +1,93 @@
...
+        require(end.commitment != 0, "Checkpoint commitment cannot be 0");
+        // TODO: once the proving incentive mechanism is in place we should reconsider this requirement because
+        // ideally we would use the proof that creates the longest chain of proven publications.
+        require(transitions[startHash] == 0, "Checkpoint already has valid transition");
+        require(start.publicationId < end.publicationId, "Start must be before end");
+        require(publicationFeed.getPublicationHash(end.publicationId) != 0, "Publication does not exist");
+
+        verifier.verifyProof(
+            publicationFeed.getPublicationHash(start.publicationId),
+            publicationFeed.getPublicationHash(end.publicationId),
```
dantaik, line 58: remove duplicate call `publicationFeed.getPublicationHash(end.publicationId)`?

src/protocol/CheckpointTracker.sol
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        _updateProvenHash();
+    }
+
+    /// @dev To limit the overhead, we cannot require a proof to find the last proven checkpoint
+    /// Instead, each proof should advance the checkpoint by a manageable increment, regardless
+    /// of which transition it proves.
+    function _updateProvenHash() internal {
+        bytes32 newProvenHash = transitions[provenHash];
```
dantaik, line 80: brain burner for me. Maybe a few test will help.

dantaik (approved):

dantaik (approved):

PR #44, opened 2025-02-24
[DRAFT] Incentivize proving via a dutch auction after the publications are known

ggonzalez94:
This PR has a proposed design for incentivizing proving via an "implicit" dutch auction that runs on a recurring time basis. This PR  builds on top of #42 .

This is a very draft PR since we decided to go back to use the ahead of time auction. I'm just pushing this so we can have it on the books in case we want to come back to the auction after the fact.
If we do, there's still a few important things to do:
- Merge the Auction and FeePool structs into one
- Match publications to period on the `CheckpointTracker`
- Deal with delayed inclusions


ggonzalez94:
Closing this in favor of #49

PR #45, opened 2025-02-24
[DRAFT] Signal Service

ernestognw:
Turned the SignalService into a library so that applications can use it to signal values. The meaning of such values depend completely on the application.

image

Also, a CommitmentSyncer is added, who's in charge of syncing the most recent commitment per chain id by using storage proofs. The only condition is that a trusted state root is required.


src/signal/state/StateSyncer.sol
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    function verifyState(uint64 chainId, uint64 blockNumber, bytes32 root, bytes[] calldata proof)
+        public
+        view
+        returns (bool valid)
+    {
+        bytes32 slot =
+            signalService().signalSlot(block.chainid.toUint64(), address(this), syncSignal(chainId, blockNumber, root));
+        return slot.verifyState(syncSignal(chainId, blockNumber, root), root, proof);
```
ernestognw, line 64: Change to verify on the signal service

src/signal/state/StateSyncer.sol
```diff
@@ -0,0 +1,82 @@
...
+        _;
+    }
+
+    constructor(address signalService_) {
+        _signalService = signalService_;
+    }
+
+    function syncSignal(uint64 chainId, uint64 blockNumber, bytes32 root)
+        public
+        view
```
ernestognw, line 30: Make it pure

nikeshnazareth:
I did the renaming and a little bit of cleanup.

I think there are still some build problems associated with the Signal Service that I am ignoring.

I'm not quite clear on what `verifyCommitment` (previously `verifyState`) achieves. As I understand it, the relevant signal will be created in `syncCommitment` (previously `syncState`) on the relevant chain, and it will be included in the Signal Service on that chain.

We're then checking that the signal exists in that chain's commitment, which I guess makes sense because it will be a record in that chain. But where do we get that chain's commitment from? Is that solved at a higher level by whoever calls this contract?

nikeshnazareth:
I modified `verifyCommitment` so that it includes both a commitment and a state root. The thought is:
- a commitment is a generic item that represents everything the rollup needs to care about at some point in time.
- at the moment we treat it like the L2 state root but it could be the L2 block hash, or a combination of L2 block hash / and embedded rollup hash / L1 state / whatever
- whoever is calling `verifyCommitment` is checking that the commitment signal is somewhere inside the state represented by the commitment. They should be responsible for retrieving the state corresponding to the commitment (which in our cases is identical to the commitment)

src/protocol/SignalService.sol
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    modifier onlyRollup() {
+        require(msg.sender == _rollup, CallerNotAuthorised());
+        _;
+    }
+
+    /// @dev Only required to be called on L1
+    function sendSignal(bytes32 value) external returns (bytes32) {
+        return value.signal();
```
LeoPatOZ, line 29: I was thinking we should definitely have an event here so that the prosper can listen to any signals being sent.

src/protocol/SignalService.sol
```diff
@@ -0,0 +1,63 @@
...
+        bytes32 value,
+        bytes[] memory accountProof,
+        bytes[] memory storageProof
+    ) external view {
+        if (accountProof.length == 0) {
+            require(_receivedSignals[LibSignal.deriveSlot(chainId, account, value)], SignalNotReceived());
+            return;
+        }
+
+        (bool valid,) = LibSignal.verifySignal(account, root, chainId, value, accountProof, storageProof);
```
dantaik, line 54: Here, we should also check root is known (verified) by consulting the rollup protocol (or the commitment syncher)

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,26 @@
...
+/// @notice Interface for the SignalService contract.
+interface ISignalService {
+    error SignalNotReceived();
+    error CallerNotAuthorised();
+
+    event SignalsReceived(bytes32[] signalSlots);
+
+    function sendSignal(bytes32 value) external returns (bytes32);
+
+    function receiveSignals(bytes32[] calldata signalSlots) external;
```
dantaik, line 14: should we also provide "isSignalSent" and "isSignalReceived" function?

src/protocol/CommitmentSyncer.sol
```diff
@@ -0,0 +1,109 @@
...
+import {LibSignal} from "../libs/LibSignal.sol";
+import {LibTrieProof} from "../libs/LibTrieProof.sol";
+
+import {ICheckpointTracker} from "./ICheckpointTracker.sol";
+import {ICommitmentSyncer} from "./ICommitmentSyncer.sol";
+
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+
+/// @dev Tracks and synchronizes commitments from different chains using their chainId.
+abstract contract CommitmentSyncer is ICommitmentSyncer {
```
dantaik, line 13: should we remove "abstract"?

src/protocol/CommitmentSyncer.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /// @dev Internal version of `syncCommitment` without access control.
+    /// Emits `CommitmentSynced` if the provided `height` is larger than `latestHeight` for `chainId`.
+    function _syncCommitment(bytes32 id, uint64 chainId, uint64 height, bytes32 commitment) internal virtual {
+        if (latestHeight(chainId) < height) {
+            _latestHeight[chainId] = height;
+            _commitment[chainId][height] = commitment;
+            // Invariant:
+            // assert(id == commitmentId(chainId, height, commitment));
+            id.signal();
```
dantaik, line 89: So now there are two SSTORE: `_commitment[chainId][height] = commitment` and `id.signal()`, can we just keep the latter?

src/protocol/CommitmentSyncer.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /// @inheritdoc ICommitmentSyncer
+    function syncCommitment(uint64 chainId, uint64 height, bytes32 commitment, bytes32 root, bytes[] calldata proof)
+        external
+        virtual
+        onlyCheckpointTracker
+    {
+        _syncCommitment(_checkCommitment(chainId, height, commitment, root, proof), chainId, height, commitment);
+    }
+
```
ernestognw, line 80:
```suggestion
    function syncCommitment(uint64 chainId, uint64 height, bytes32 root, bytes[] calldata proof)
        external
        virtual
        onlyCheckpointTracker
    {
        _syncCommitment(_checkCommitment(chainId, height, commitment, root, proof), chainId, height, root);
    }

```

@nikeshnazareth, was this your suggestion?

PR #49, opened 2025-02-25
Ahead of time auction for prover incentives

### LINKED ISSUE #15: Incentivize proving and set bonds
In the current version of Taiko protocol, proposers are on the hook to prove their on batches and deposit a `livenessBond` to make sure they do so(or if they don't do it in time a prover can come and claim a part of their bond).
Under this new design we don't have a specific `proposeBatches` function in the Inbox, and the `DataFeed` is intended to be used across multiple rollups.
We would like to have a design where provers are incentivized and compensated to prove multiple publications(regardless of who proposed them), instead of each proposer having to prove their own(which is less efficient).
We would also like a design that is "modular", where the protocol is agnostic to how these incentives work, and each rollup can "extend" this by adding their own mechanism(e.g. some rollup may decide to subsidize proving with token emissions, while others may prefer a prover marketplace that gets the best price per proof)

ggonzalez94:
Fixes #15 

This PR includes the design for an implicit ahead of time auction and all the logic and accounting necessary for taking proposers' fees and paying and slashing provers. It implements the design exposed [here](https://github.com/OpenZeppelin/minimal-rollup/pull/62).

## Some design assumptions

- The `ProverManager` contract implements two interfaces, the `IProposerFees` and the `IProverManager` interfaces. It is called by the Inbox contract to pay the fees for each proposal and calls the `CheckPointTracker` when proving publications.
- When deployed an initial prover is set. This simplifies some of the logic, as we can assume there's always an active period(making the implementation more efficient), and proposers can directly submit publications. This shouldn't be an issue, as the rollup operator can set themselves as an initial prover, with a high enough fee so that other participants can outbid them relatively easily.
- This design assumes only sequential proving is allowed, so #56 should be implemented. Every proof starts from the latest known checkpoint and advances the chain, there's no need to track transitions. This makes the market design easier since we don't need to keep track of separate proving deadlines, or non consecutive periods.

## What this PR does not include

These is a list of things this PR does not include to keep the scope manageable. These will be submitted as separated PRs or tracked as separate issues.
- Gas optimizations(will be submitted as a separate PR). Particularly packing the `Period` struct into only two slots yields ~20% gas improvements.
- Logic for a permissioned actor(e.g. a DAO) to change some of the parameters(`livenessWindow`, `livenessBond`, etc.)
- How to price delayed publications(since we still need to decide if we will price them differently). The contract has the parameters necessary to deal with delayed publications, but would still need some adjustments. Tracked under #68 
- Logic to work with ERC20 tokens. Tracked under #51


src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+            emit Deposit(proposer, msg.value);
+        }
+
+        uint256 _currentPeriod = currentPeriodId;
+
+        uint256 periodEnd = periods[_currentPeriod].end;
+        if (periodEnd != 0 && block.timestamp > periodEnd) {
+            // Advance to the next period
+            currentPeriodId++;
+            _currentPeriod++;
```
ernestognw, line 152: May save a couple of PUSH/POP/DUP

```suggestion
             currentPeriodId = ++_currentPeriod;
```

ernestognw (commented):
Looking good, will continue reviewing later

src/protocol/CheckpointTracker.sol
```diff
@@ -29,20 +30,24 @@ contract CheckpointTracker is ICheckpointTracker {
     /// @param _genesis the checkpoint commitment describing the initial state of the rollup
     /// @param _publicationFeed the input data source that updates the state of this rollup
     /// @param _verifier a contract that can verify the validity of a transition from one checkpoint to another
```
LeoPatOZ, line 32:
```suggestion
    /// @param _verifier a contract that can verify the validity of a transition from one checkpoint to another
    /// @param _proverManager contract responsible for managing the prover auction
```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+        livenessBond = _config.livenessBond;
+        evictorIncentivePercentage = _config.evictorIncentivePercentage;
+        burnedStakePercentage = _config.burnedStakePercentage;
+        inbox = _inbox;
+        checkpointTracker = ICheckpointTracker(_checkpointTracker);
+        publicationFeed = IPublicationFeed(_publicationFeed);
+
+        // Initialize the first period with a known prover and a set fee
+        require(msg.value >= _config.livenessBond, "Insufficient balance for liveness bond");
+        periods[0].prover = _initialProover;
```
LeoPatOZ, line 112:
```suggestion
        periods[0].prover = _initialProver;
```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+        provingDeadline = _config.provingDeadline;
+        livenessBond = _config.livenessBond;
+        evictorIncentivePercentage = _config.evictorIncentivePercentage;
+        burnedStakePercentage = _config.burnedStakePercentage;
+        inbox = _inbox;
+        checkpointTracker = ICheckpointTracker(_checkpointTracker);
+        publicationFeed = IPublicationFeed(_publicationFeed);
+
+        // Initialize the first period with a known prover and a set fee
+        require(msg.value >= _config.livenessBond, "Insufficient balance for liveness bond");
```
LeoPatOZ, line 111: Do we not need to add this deposit to the provers 'balance'? 

```
balances[_initialProver] += amount;
```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+        uint256 periodEnd = periods[_currentPeriod].end;
+        if (periodEnd != 0 && block.timestamp > periodEnd) {
+            // Advance to the next period
+            currentPeriodId++;
+            _currentPeriod++;
+            emit NewPeriod(_currentPeriod);
+        }
+
+        uint256 requiredFee = periods[_currentPeriod].fee;
+
```
LeoPatOZ, line 157: why do we not require ?
```solidity
require(balances[proposer] >= requiredFee, "Insufficient balance")
```

is that redundant

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+        uint256 end; // the end of the period(this may happen because the prover exits, is evicted or outbid)
+        uint256 deadline; // the time by which the prover needs to submit a proof
+    }
+
+    /// @dev This struct is necessary to pass it to the constructor and avoid stack too deep errors
+    /// When some values in the contract stop being immutable, we may change this to be more efficient
+    struct ProverManagerConfig {
+        uint256 minUndercutPercentage;
+        uint256 livenessWindow;
+        uint256 succesionDelay;
```
LeoPatOZ, line 24:
```suggestion
        uint256 successionDelay;
```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+    /// @dev This is used to prevent gas wars where the new prover undercuts the current prover by just a few wei
+    uint256 public immutable minUndercutPercentage;
+    /// @notice The time window after which a publication is considered old enough and if the prover hasn't poven it yet
+    /// can be evicted
+    uint256 public immutable livenessWindow;
+    /// @notice The delay after which the next prover becomes active
+    /// @dev The reason we don't allow this to happen immediately is so that:
+    /// 1. Other provers can bid for the role
+    /// 2. Ensure the current prover window is not too short
+    uint256 public immutable succesionDelay;
```
LeoPatOZ, line 49:
```suggestion
    uint256 public immutable successionDelay;
```

LeoPatOZ:
More of a question rather than a suggestion but was wondering if we could merge some of the params. 
Specifically `successionDelay` and `exitDelay`.  Unless theres a really strong reason to keep these two conifgurable perhaps we this could simply be:
```solidity
/// @notice The delay for prover transitions, including succession and exit delays
uint256 public immutable proverTransitionDelay;
```
could reduce some complexity

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+        _validateNextPublicationHeader(nextPublicationHeader, end.publicationId + 1, periodEnd);
+
+        checkpointTracker.proveTransition(start, end, proof);
+
+        // Distribute the funds
+        uint256 _livenessBond = period.stake;
+        uint256 accumulatedFees = period.accumulatedFees;
+        uint256 newProverFees = period.fee * numPubs;
+
+        // Pay the designed prover for the work they already did
```
LeoPatOZ, line 331:
```suggestion
        // Pay the designated prover for the work they already did
```
?

src/protocol/taiko_alethia/ProverManager.sol
ernestognw: Yes, line 159 will revert in case of overflow. If `requiredFee > balances[proposer]`, it will [panic with code `0x11`](https://github.com/ethereum/solidity/blob/v0.8.24/libsolutil/ErrorCodes.h#L29)

LeoPatOZ: maybe still nice to explicitly panic though idm

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+    /// conditions.
+    function bid(uint256 offeredFee) external {
+        uint256 _livenessBond = livenessBond;
+        require(balances[msg.sender] >= _livenessBond, "Insufficient balance for liveness bond");
+
+        uint256 currentPeriod = currentPeriodId;
+        Period storage _currentPeriod = periods[currentPeriod];
+        Period storage _nextPeriod = periods[currentPeriod + 1];
+        uint256 requiredMaxFee;
+        if (_currentPeriod.end == 0) {
```
ernestognw, line 175: I think we should wrap `_currentPeriod.end == 0` in an `_isPeriodActive(uint256 end)`  internal function. If so:

```suggestion
        if (_isPeriodActive(_currentPeriod.end)) {
```

The core idea is that the contract API should allow developers customizing the rollup to bring their own definitions of what each piece of the contract means.

It's also true that this contract is specific to Taiko. Maybe is it worth discussing the level of customizability we want to allow and where?

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+            uint256 currentFee = _currentPeriod.fee;
+            requiredMaxFee = currentFee - calculatePercentage(currentFee, minUndercutPercentage);
+            require(offeredFee <= requiredMaxFee, "Offered fee not low enough");
+
+            uint256 periodEnd = block.timestamp + succesionDelay;
+            _currentPeriod.end = periodEnd;
+            _currentPeriod.deadline = periodEnd + provingDeadline;
+        } else {
+            address _nextProverAddress = _nextPeriod.prover;
+            if (_nextProverAddress != address(0)) {
```
ernestognw, line 186: Following the same logic of the my previous comment, there should be an abstraction for this, something like `_isBidded(address nextProver)` or similar. And then:

```suggestion
            if (_isBidded(_nextPeriod.prover)) {
```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+        return periods[periodId];
+    }
+
+    /// @dev Calculates the percentage of a given numerator scaling up to avoid precision loss
+    /// @param amount The number to calculate the percentage of
+    /// @param bps The percentage expressed in basis points(https://muens.io/solidity-percentages)
+    /// @return _ The calculated percentage of the given numerator
+    function calculatePercentage(uint256 amount, uint256 bps) private pure returns (uint256) {
+        require((amount * bps) >= 10_000);
+        return amount * bps / 10_000;
```
ernestognw, line 353: Perhaps the decimals is another abstraction we want?

```suggestion
        return amount * bps / feeDecimals();
```

As part of the interface, it would be a single virtual function to override the decimals of all the fee logic in the contract

ernestognw (commented):
I like the direction this is taking! I'd like to discuss more about creating overrideable abstractions for developers. I understand the idea of the ProverManager is taiko specific, but perhaps Taiko is a specialization of an abstract contract?

I think the idea is worth exploring if we want to have a semi-standardized approach about customizing ProverManager logic

ernestognw:
> Was wondering if we could merge some of the params. Specifically `successionDelay` and `exitDelay`. Unless theres a really strong reason to keep these two configurable perhaps this could simply be:
> 
> ```solidity
> /// @notice The delay for prover transitions, including succession and exit delays
> uint256 public immutable proverTransitionDelay;
> ```
> 
> could reduce some complexity

This is another useful reason to have an abstract contract. We as the library designers can't either recommend or expect that both delays are going to be the same, but if we make an abstract `ProvingManager` with `successionDelay()` and `exitDelay()` overridable public functions, then the implementation can decide whether to use the same slot or not.

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+        uint256 minUndercutPercentage;
+        uint256 livenessWindow;
+        uint256 succesionDelay;
+        uint256 exitDelay;
+        uint256 delayedFeeMultiplier;
+        uint256 provingDeadline;
+        uint256 livenessBond;
+        uint256 evictorIncentivePercentage;
+        uint256 burnedStakePercentage;
+    }
```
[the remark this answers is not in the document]
ernestognw: Both can achieve very similar performance. In theory, you can use the struct in-storage without allocating memory, which is the only reason I'd think it'd be worth optimize for.

My closest experience with something like this is the OZ Governor, where there's an interface but the internal proposal is stored as a struct. Generally the struct goes in the most derived contract unless the interface wants to enforce it.

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+
+contract ProverManager is IProposerFees, IProverManager {
+    struct Period {
+        address prover;
+        uint256 stake; // stake the prover locked to register
+        uint256 accumulatedFees; // the fees accumulated by proposers' publications for this period
+        uint256 fee; // per-publication fee (in wei)
+        uint256 end; // the end of the period(this may happen because the prover exits, is evicted or outbid)
+        uint256 deadline; // the time by which the prover needs to submit a proof
+    }
```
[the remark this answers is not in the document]
ggonzalez94: Agreed, and we can pack even further if we are willing to make the balances smaller. For example a `uint88`is enough to hold all the TAIKO circulating supply and much more than any ETH this contract might hold, but I'll push those optimizations for a next PR, since we still have a few changes that will affect storage layout(support for ERC20, decide if some of the configuration parameters should actually be storage variables, how to deal with delayed publications)

src/protocol/taiko_alethia/ProverManager.sol
ggonzalez94: No, because it directly goes to the staking amount. One thing we could do is require that `msg.value == _config.livenessBond`. I don't think there's a reason in the constructor you may want to send more than that, but it doesn't break anything since they prover would still get that stake back after completing their period(the only thing is it does put them at a risk of getting slashed by a bigger amount)

src/protocol/taiko_alethia/ProverManager.sol
ggonzalez94: The only reason to do it explicitely is because of a nicer error message that wallets can interpret, but it does consume gas. That's why I decided to rely on the implicit check

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+    /// @dev This value needs to be expressed in basis points
+    uint256 public immutable burnedStakePercentage;
+
+    /// @notice Common balances for proposers and provers
+    mapping(address user => uint256 balance) public balances;
+    /// @notice Periods represent proving windows
+    /// @dev Most of the time we are dealing with the current period or next period (bids for the next period),
+    /// but we need periods in the past to track publications that still need to be proven after the prover is
+    /// evicted or exits
+    mapping(uint256 periodId => Period period) private periods;
```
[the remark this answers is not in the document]
ggonzalez94: Seems like the kind of things we should put in the minimal standards 😉

src/protocol/CheckpointTracker.sol
```diff
@@ -29,20 +30,24 @@ contract CheckpointTracker is ICheckpointTracker {
...
         Checkpoint memory genesisCheckpoint = Checkpoint({publicationId: 0, commitment: _genesis});
         provenHash = keccak256(abi.encode(genesisCheckpoint));
         emit CheckpointUpdated(provenHash);
     }
 
     /// @inheritdoc ICheckpointTracker
     function proveTransition(Checkpoint calldata start, Checkpoint calldata end, bytes calldata proof) external {
+        if (proverManager != address(0)) {
+            require(msg.sender == proverManager, "Only the prover manager can call this function");
+        }
```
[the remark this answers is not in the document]
ggonzalez94: Yeah, it should be the same. But do you think it is more readable?
I like that in the case of the explicit outer if you can read:
If there is a proverManager, then...

Anyways it is a small detail and if you have a strong preference towards the `require`I'm ok with it

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -65,6 +69,7 @@ contract TaikoInbox {
         attributes[LAST_PUBLICATION] = abi.encode(_lastPublicationId);
         attributes[BLOB_REFERENCE] = abi.encode(blobRefRegistry.getRef(_buildBlobIndices(nBlobs)));
 
+        proposerFees.payPublicationFee{value: msg.value}(msg.sender, false);
```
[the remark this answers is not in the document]
ggonzalez94: Good catch, we're still a bit uncertain how we will deal with delayed publications. One possible approach I'm think:
- We have a `getFee`function on the `ProverManager` that returns a tuple with the value for regular and delayed publications. For now it can return the same value in both.
- The Inbox calls this function before calling  `payPublicationFee`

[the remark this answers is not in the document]
ggonzalez94: Went ahead with this approach. For now the `ProverManager` returns the same value for both regular and delayed publications and we have a TODO for when we decide how to price them.
This is a function that when we make the abstract contract, implementers can overwrite to decide how they price this

ggonzalez94:
> I'd like to discuss more about creating overrideable abstractions for developers

After our call and thinking about it more I'm very much in favor of the abstract contract and all these parameters being overridable functions. I created an issue and assigned it to you @ernestognw #72 , but I'm happy to work on it together

src/protocol/taiko_alethia/ProverManager.sol
ggonzalez94: Do you think this flexibility is useful? I believe `bps` is a standard way of representing percentages and if you change the denominator then you have to also be careful with the numerator. If someone changes the `feeDecimals()` implementation an still uses basis points to represent percentages they will shoot themselves in the foot

ggonzalez94:
> I dont know if its in scope for this PR but maybe we should expose a getCurrentRequiredFee which returns the fee for this period. This could be useful in the delayed inclusion flow where we might force the user to pay the provers fee (or a % of the required fee)

Yes, good call. I implemented this as part of the `IProposerFees` interface that the `ProverManager` inherits from. For now it returns the same value for both regular and delayed publications, but once we decide how to deal with delayed we can change that

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,405 @@
...
+
+        emit Deposit(msg.sender, msg.value);
+    }
+
+    /// @notice Withdraw available(unlocked) funds.
+    /// @param amount The amount to withdraw
+    function withdraw(uint256 amount) external {
+        balances[msg.sender] -= amount;
+
+        (bool ok,) = payable(msg.sender).call{value: amount}("");
```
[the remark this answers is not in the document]
ggonzalez94: Well... I thought the compiler would be smart enough to solve this(especially if you set the `optimizer_runs` to a higher value), but I was wrong

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -78,6 +84,7 @@ contract TaikoInbox is IInbox {
             attributes[LAST_PUBLICATION] = abi.encode(_lastPublicationId);
             attributes[BLOB_REFERENCE] = abi.encode(inclusions[i]);
 
+            proposerFees.payPublicationFee{value: delayedPublicationFee}(msg.sender, true);
```
nikeshnazareth, line 87: should the value come from the person who originally made the delayed publication? Otherwise, it seems we're forcing some random proposer to cover the cost of a delayed publication that they didn't choose

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,436 @@
...
+    function withdraw(uint256 amount) external {
+        balances[msg.sender] -= amount;
+
+        address to = msg.sender;
+        bool ok;
+        // Using assembly to avoid memory allocation costs; only the call's success matters to ensure funds are sent.
+        assembly ("memory-safe") {
+            ok := call(gas(), to, amount, 0, 0, 0, 0)
+        }
+
```
nikeshnazareth, line 126:
```suggestion
require(ok, "Withdraw failed");

```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,436 @@
...
+        _nextPeriod.stake = _livenessBond;
+        balances[msg.sender] -= _livenessBond;
+
+        emit ProverOffer(msg.sender, currentPeriod + 1, offeredFee, _livenessBond);
+    }
+
+    /// @inheritdoc IProverManager
+    /// @dev This can be called by anyone, and they get `evictorIncentivePercentage` of the liveness bond as an
+    /// incentive.
+    function evictProver(uint256 publicationId, IPublicationFeed.PublicationHeader calldata publicationHeader)
```
nikeshnazareth, line 203: We need a check to ensure the publication hasn't already been proven. As it stands you can pick any publication in the past and evict the current prover because of it.

src/protocol/CheckpointTracker.sol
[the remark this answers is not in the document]
nikeshnazareth: I don't think it is equivalent. The equivalent statement would be

```solidity
require(proverManger == address(0) || msg.sender == proverManager, "...")
```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,446 @@
...
+        uint256 periodEnd
+    ) private view {
+        require(nextPub.id == expectedId, "This is not the next publication");
+        require(nextPub.timestamp > periodEnd, "Next publication is not after the period end");
+        require(publicationFeed.validateHeader(nextPub, nextPub.id), "Publication hash does not match");
+    }
+
+    /// @dev Validates a chain of publication headers.
+    /// @param headers The chain of publication headers to validate
+    function _validatePublicationChain(IPublicationFeed.PublicationHeader[] memory headers) private view {
```
nikeshnazareth, line 410: I don't think this is correct. The reason you're passing a chain at all (instead of just comparing the publication id values of start/end) is that you're trying to count how many publications are relevant to the Taiko inbox. This means we need to skip publications that didn't come from that inbox, and we can check that it's still a valid chain by using the `LAST_PUBLICATION` attribute

nikeshnazareth: Instead of trying to disentangle all this on-chain, would it make sense to pass a `numPublications` parameter to the prover, so we can validate it inside the proof?

nikeshnazareth: Alternatively, we should just check the `publisher` field of all the publications they want to prove, to show they came from the Taiko Inbox (which we should do anyway). We don't have to validate that the chain of publications is correct, just that they fall within the start/end checkpoints. After the `proveTransition` call:
- the caller has advanced the last proven checkpoint over an arbitrary number of publications, some of which came from the Taiko inbox
- they get paid for all the publications in that range that they prove came from the inbox

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -346,14 +345,28 @@ contract ProverManager is IProposerFees, IProverManager {
...
         uint256 newProverFees = period.fee * numPubs;
 
         // Pay the designated prover for the work they already did
-        balances[period.prover] += accumulatedFees - newProverFees;
+        balances[period.prover] += period.accumulatedFees - newProverFees;
 
         // Compensate the new prover with fees and a portion of the stake
         _distributeStakeAndFees(period.stake, newProverFees, msg.sender);
+
+        period.accumulatedFees = period.stake = 0;
```
ggonzalez94, line 356: I think this is fine for now, but once we decide what to do with the "burned" token we might want to change this so that the stake for the period is not zeroed out or we keep track of that number somewhere else

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -465,4 +478,29 @@ contract ProverManager is IProposerFees, IProverManager {
...
+
+    function _isClosed(uint256 periodEnd, ICheckpointTracker.Checkpoint calldata lastProven, bytes calldata headerBytes)
+        private
+        returns (bool)
+    {
+        bytes32 lastProvenHash = keccak256(abi.encode(lastProven));
+        require(lastProvenHash == checkpointTracker.provenHash(), "Incorrect lastProven checkpoint");
+
+        // Case 1: all publications are proven and the period is over
+        if (publicationFeed.getNextPublicationId() == lastProven.publicationId + 1 && block.timestamp > periodEnd) {
```
ggonzalez94, line 490: @nikeshnazareth do you think this check is really necessary? And if so I don't think it should be the first one.
- I assume we should always have new periods and publications, so things should be covered always by case 2. If you have a closed period which you didn't claim your rewards you just have to wait until there's a new proven publication with an id higher than yours.
- The only scenario I can think of where this is not true is if the chain stalls and there are no more publications, but even then people might keep adding things to the feed and this will never be true

ggonzalez94 (commented):
I think we can get rid of the first check in the new `_isClosed` function

src/protocol/taiko_alethia/ProverManager.sol
ggonzalez94: Thanks for tackling this, great catch

src/protocol/IProposerFees.sol
```diff
@@ -0,0 +1,15 @@
...
+    /// @notice Proposers have to pay a fee for each publication they want to get proven. This should be called only by
+    /// the Inbox contract.
+    /// @param proposer The address of the proposer
+    /// @param isDelayed Whether the publication is a delayed publication
+    function payPublicationFee(address proposer, bool isDelayed) external payable;
+
+    /// @notice Returns the current fees for a period
+    /// @return fee The fee for a regular publication
+    /// @return delayedFee The fee for a delayed publication
+    function getCurrentFees() external view returns (uint256 fee, uint256 delayedFee);
```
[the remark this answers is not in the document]
ggonzalez94: Do you think we should support getting fees from other periods? Right now on-chain we don't have a use case that requires knowing the fees from a different period and off-chain actors can always get that from events.
The reason this function currently gets current fees instead of taking a period as an argument is that the `Inbox`is unaware of the concept of periods. Instead it would have to ask for the current period and then pass it as a parameter

src/protocol/IProverManager.sol
```diff
@@ -0,0 +1,60 @@
...
+        bytes calldata proof,
+        uint256 periodId
+    ) external;
+
+    /// @notice Called by a prover when the originally assigned prover is passed its deadline
+    /// @dev This function should slash the prover and distribute their stake to compensate the new prover
+    /// @param start The initial checkpoint before the transition
+    /// @param end The final checkpoint after the transition
+    /// @param publicationHeadersToProve The chain of publication headers to prove
+    /// @param nextPublicationHeader The next publication header. It should be after the period end
```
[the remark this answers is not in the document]
ggonzalez94: Maybe this is a dumb questions, but when would this happen? The only scenario I could think is when the chain stalls, but even in this unlikely scenario(that would still require an upgrade) as long as someone is posting to the common feed you would be able to get a next publication

src/protocol/taiko_alethia/ProverManager.sol
ggonzalez94: You are correct, for some reason I thought the `prevHash` pointed to the previous publication from the rollup, but it is actually just the previous publication from the feed.

> Instead of trying to disentangle all this on-chain, would it make sense to pass a numPublications parameter to the prover, so we can validate it inside the proof?

This seems like the most efficient option, would it be possible to do all of this inside the prover and not have to verify that the publications are actually from the same rollup on-chain? If so the parameters we pass to `proveClosedPeriod` would be very similar to what we pass to `proveOpenPeriod`

src/protocol/IProverManager.sol
```diff
@@ -0,0 +1,92 @@
...
+    /// This function allows anyone to start a new period with a new fee. Note that the new prover will still need to
+    // ensure the previous publications are proven before proving their own. Nevertheless, we start a new period so
+    // they cannot be evicted based on publications that occurred before they agreed to prove
+    function claimProvingVacancy(uint256 fee) external;
+
+    /// @notice Evicts a prover that has been inactive, marking the prover for slashing
+    /// @param publicationId The publication id that the caller is claiming is too old and hasn't been proven
+    /// @param publicationHeader The publication header that the caller is claiming is too old and hasn't been proven
+    /// @param lastProven The last proven checkpoint, which is used to show the publicationHeader is not proven
+    /// TODO: we should save the actual checkpoint (not the hash) in CheckpointTracker so we can query it directly
```
ggonzalez94, line 31: Is this so that the checks in the `ProverManager` become easier(instead of having to pass the actual checkpoint and hashing it)? If that's the case while I agree it will be cleaner, I still think the storage overhead of storing the entire Checkpoint probably outweighs the cost of calldata and the hashing and comparison

nikeshnazareth: Yes, to make the ProverManager simpler, and also to avoid potential race-conditions (eg. if someone tries to call this function or `finalizeClosedPeriod` and the `provenHash` changes, the calls will revert).

My intuition is that the checkpoint hash made sense when we were saving several transitions and didn't want to double the storage, but in this case doubling the storage just means one extra slot. Still, no strong opinions

ernestognw (commented):
Haven't finished reviewing, I'll continue later. For now, my most constructive feedback is that the contract requires iteration instead of patching comments with extra variables or functions.

If I'm correct, we must be able to model the whole mechanism with very few variables. I'm still not sure (have thought about this a lot), but I think a good approach to simplify it and keep iterating is trying to come up with a library to encapsulate the mechanics of the `Period`.

Once that's done, the implementation of the ProvingMarket becomes more trivial since the abstractions are more concrete. Instead, we're now using variables to represent intermediate states so it's difficult to reason about how all these variables work together.

One approach to encapsulating the period logic would be:

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.28;

import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";

/// @dev A bidded period is a time window in which a prover can post proofs for state transitions.
///
/// A period is economically protected by a liveness bond the prover has to stake. Its purpose
/// is to disincentivize the prover from not posting a proof.
///
/// Each period has the following properties:
///
/// - `prover`: The address of the prover with the rights to post proofs during the period if active.
/// - `start`: Timestamp when the period becomes active. 0 means the period is not scheduled.
/// - `deadline`: Timestamp at which the prover becomes evictable if no proof is posted.
/// - `end`: Timestamp when the period ends. 0 means the period is still active.
/// - `stake`: The amount of ETH the prover has to deposit as a liveness bond.
/// - `accumulatedFee`: The total amount of fees accumulated in the period.
/// - `offeredFee`: The fee the prover is willing to charge for proving each publication.
library ProvingPeriods {
    struct Period {
        // slot 0
        address prover;
        uint48 start;
        uint48 deadline;
        // slot 1
        uint48 end;
        uint80 stake; // Up to ~1,208,925.688 ETH
        uint64 accumulatedFee; // Up to ~18.44 ETH
        uint64 offeredFee; // Up to ~18.44 ETH
    }

    /// @notice The period has at least one bid.
    function isBidded(Period storage period) internal view returns (bool) {
        return period.prover != address(0);
    }

    /// @notice The period is active (i.e. right to post proofs are for the registered prover).
    function isActive(Period storage period) internal view returns (bool) {
        return isStarted(period) && !isEnded(period);
    }

    /// @notice The period has started.
    function isStarted(Period storage period) internal view returns (bool) {
        return period.start != 0 && SafeCast.toUint48(block.timestamp) >= period.start;
    }

    /// @notice The period has ended.
    function isEnded(Period storage period) internal view returns (bool) {
        return period.end != 0 && SafeCast.toUint48(block.timestamp) >= period.end;
    }

    /// @notice The current prover can be evicted.
    function isEvictable(Period storage period) internal view returns (bool) {
        return period.deadline != 0 && SafeCast.toUint48(block.timestamp) >= period.deadline;
    }

    /// @notice Sets the moment the next period starts with a deadline after the prover can be evicted if no proof is
    /// posted.
    function scheduleStartAfterDelayWithDeadline(Period storage period, uint48 delay, uint48 provingDeadline)
        internal
    {
        uint48 start = SafeCast.toUint48(block.timestamp) + delay;
        period.start = start;
        period.deadline = start + provingDeadline;
    }

    /// @notice Sets the moment the current period ends after a delay.
    function scheduleEnd(Period storage period) internal returns (uint256) {
        uint48 periodEnd = SafeCast.toUint48(block.timestamp);
        period.end = periodEnd;
        // Respect the previous deadline if set
        // uint48 cast is safe because both values are uint48
        uint48 newDeadline = uint48(Math.min(period.deadline, periodEnd));
        period.deadline = newDeadline == 0 ? periodEnd : newDeadline;
        return periodEnd;
    }
}
```

I'm pretty sure the implementation could be substantially simplified, and we should aim for an indeed _minimal_ version with customization instructions. These instructions must rely on language features like overriding functions as opposed to referencing external contracts.

src/protocol/taiko_alethia/TaikoInbox.sol
ggonzalez94: Yes, this is not yet implemented. Tracking it [here](https://github.com/OpenZeppelin/minimal-rollup/issues/68) and plan to implement it once we merge the initial delayed inclusion.

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,509 @@
...
+    /// @notice The minimum stake required to be a prover
+    /// @dev This should be enough to cover the cost of a new prover if the current prover becomes inactive
+    uint256 public immutable livenessBond;
+    /// @notice The percentage of the liveness bond that the evictor gets as an incentive
+    /// @dev This value needs to be expressed in basis points (4 decimals)
+    uint256 public immutable evictorIncentivePercentage;
+    /// @notice The percentage of the liveness bond (at the moment of the slashing) that is burned when a prover is
+    /// slashed
+    /// @dev This value needs to be expressed in basis points (4 decimals)
+    uint256 public immutable burnedStakePercentage;
```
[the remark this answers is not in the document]
ggonzalez94: Agree, tracked in #72 and will tackle it next after we finish the initial implementation

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,509 @@
...
+
+    // -- Configuration parameters --
+    /// @notice The minimum percentage by which the new bid has to be lower than the current best value
+    /// @dev This value needs to be expressed in basis points (4 decimal places)
+    /// @dev This is used to prevent gas wars where the new prover undercuts the current prover by just a few wei
+    uint256 public immutable minUndercutPercentage;
+    /// @notice The time window after which a publication is considered old enough and if the prover hasn't proven it
+    /// yet
+    /// can be evicted
+    uint256 public immutable livenessWindow;
```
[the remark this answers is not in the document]
ggonzalez94: `livenessWindow` is how much time you have between any publication and when you submit a proof(it is used to evict provers if they don't publish anything for some period of time).
`provingWindow` is the amount of time a prover has to close their entire period after they exit or are outbid.

So they can(and I imagine will usually be) different values, you might want to keep the livenessWindow tighter to ensure there are more frequent checkpoints, but the provingDeadline might actually be longer to ensure a prover has enough time to close it since it is a potentially heavier operation to close the entire period.
The two variables that I do think could potentially be merged are `successionDelay` and `exitDelay`, since they both refer to a window before the next period starts. However this will become irrelevant once we make all this annoying constants internal overridable functions

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,509 @@
...
+        uint256 evictorIncentivePercentage;
+        uint256 burnedStakePercentage;
+    }
+
+    address public immutable inbox;
+    ICheckpointTracker public immutable checkpointTracker;
+    IPublicationFeed public immutable publicationFeed;
+
+    // -- Configuration parameters --
+    /// @notice The minimum percentage by which the new bid has to be lower than the current best value
```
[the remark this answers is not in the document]
ggonzalez94: Noted -  -I'm keeping the `notice` for now since today they are public, but this will be changed to internal functions that will just have a `@dev`

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,509 @@
...
+    /// @dev This value needs to be expressed in basis points (4 decimals)
+    uint256 public immutable burnedStakePercentage;
+
+    /// @notice Common balances for proposers and provers
+    mapping(address user => uint256 balance) public balances;
+    /// @notice Periods represent proving windows
+    /// @dev Most of the time we are dealing with the current period or next period (bids for the next period),
+    /// but we need periods in the past to track publications that still need to be proven after the prover is
+    /// evicted or exits
+    mapping(uint256 periodId => Period) private _periods;
```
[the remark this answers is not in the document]
ggonzalez94: Agree that given the access pattern we could use an array, but not sure if it is worth it. I did a quick dirty replacement(probably can be optimized) and the extra `push` do add some gas overhead. I'll think more about it, but for now it seems that we are better with a mapping
image

src/protocol/IProverManager.sol
ggonzalez94: I ended up changing it and storing the entire CheckPoint vs just the hash. I was slightly worried about gas usage, but while `proveTransition` in the `CheckPointTracker` is more expensive the simpler checks seem to make up for that and the gas cost of `prove` is very similar. `evictProver` is more expensive, but since this should be called very rarely, we should be fine

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,509 @@
...
+    event NewPeriod(uint256 period);
+
+    constructor(
+        address _inbox,
+        address _checkpointTracker,
+        address _publicationFeed,
+        address _initialProver,
+        uint256 _initialFee,
+        ProverManagerConfig memory _config
+    ) payable {
```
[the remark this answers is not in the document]
ggonzalez94: Should also be refactored with the abstract implementation

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,509 @@
...
+    /// @inheritdoc IProposerFees
+    /// @dev This function advances to the next period if the current period has ended.
+    // TODO: deal with delayed publications
+    function payPublicationFee(address proposer, bool isDelayed) external payable {
+        require(msg.sender == inbox, "Only the Inbox contract can call this function");
+
+        // Accept additional deposit if sent
+        if (msg.value > 0) {
+            _deposit(proposer, msg.value);
+        }
```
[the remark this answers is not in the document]
ggonzalez94: I don't think it hurts though.... we spend 15 gas on the check if no eth is sent, and gives some flexibility to proposers to deposit it along with their tx. But agree it might be worth be simplifying it

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,357 @@
...
+        uint256 _livenessBond = livenessBond;
+        _updatePeriod(_nextPeriod, msg.sender, offeredFee, _livenessBond);
+
+        emit ProverOffer(msg.sender, currentPeriod + 1, offeredFee, _livenessBond);
+    }
+
+    /// @inheritdoc IProverManager
+    /// @dev This can be called by anyone, and they get `evictorIncentivePercentage` of the liveness bond as an
+    /// incentive.
+    function evictProver(IPublicationFeed.PublicationHeader calldata publicationHeader) external {
```
xiaodino, line 200: Looking at the eviction logic, it seems that when a prover is evicted (or slashed for inactivity), their staked liveness bond remains in the contract indefinitely, with no immediate way to reclaim or redistribute it. In `evictProver`, after paying out the evictor incentive, the remaining stake stays in `period.stake`.

If no one finalizes the period (i.e., no call to `finalizeClosedPeriod`), the leftover bond just sits locked. Is this the intended behavior? It effectively means the slashed prover’s funds are stuck unless a new prover completes the job. 

Should there be a fallback mechanism to handle abandoned stake in cases where no one follows up?

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,357 @@
...
+        require(publicationHeader.id > lastProven.publicationId, "Publication has been proven");
+
+        uint256 periodEnd = block.timestamp + exitDelay;
+        // We use this to mark the prover as evicted
+        _finalizePeriod(period, periodEnd, periodEnd);
+
+        // Reward the evictor and slash the prover
+        uint256 evictorIncentive = _calculatePercentage(period.stake, evictorIncentivePercentage);
+        balances[msg.sender] += evictorIncentive;
+        period.stake -= evictorIncentive;
```
xiaodino, line 219: What happens if `period.stake` is already zero at the time of this call?

Should we explicitly guard against `period.stake` being zero (or generally ensure `period.stake` >= `evictorIncentive`) before this subtraction? An explicit require or condition could make the intent clearer and prevent a confusing underflow revert. It might also improve clarity by signaling that an eviction is only valid when there’s stake to slash. 

For instance, if no prover is active or the stake is already depleted, perhaps the eviction should be disallowed upfront (ensuring the function is only called on an “active” period with a positive stake).

In summary, do we need to handle the case where `period.stake` is 0 more gracefully? Or is there already a condition ensuring this can’t happen (e.g. a check that the period is active or a minimum stake requirement)?

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,357 @@
...
+        _finalizePeriod(period, periodEnd, _provingDeadline);
+
+        emit ProverExited(_prover, periodEnd, _provingDeadline);
+    }
+
+    /// @inheritdoc IProverManager
+    function claimProvingVacancy(uint256 fee) external {
+        uint256 periodId = currentPeriodId;
+        Period storage period = _periods[periodId];
+        require(period.prover == address(0) && period.end == 0, "No proving vacancy");
```
xiaodino, line 244: I have a question about the flow after a prover exits or is evicted. 

When `exit()` is called by the current prover, you set an `exitDelay` and `provingDeadline` but the `period.prover` remains the same (the exiting prover) in storage​.

Similarly, after `evictProver`, the `period.prover` still points to the evicted prover (since `_finalizePeriod` only sets timestamps and doesn’t nullify the prover). 

Yet the `claimProvingVacancy()` function requires `period.prover == address(0)` for a new prover to claim the role​.

 In what scenario would `period.prover` actually be 0 to allow `claimProvingVacancy`? Should the contract be setting `period.prover = address(0)` when a prover exits or is evicted to signal an open slot?

nikeshnazareth:
Should we unite `livenessWindow` and `provingWindow`?

- I think we want `provingWindow` to be at least as large as `livenessWindow`, because otherwise a publication can come in on the `end` timestamp, and you will only have `provingWindow` to prove it.
- On the other hand, there doesn't seem to be a reason for `provingWindow` to be larger than `livenessWindow`.

Similarly, is there any reason that `exitDelay` and `successionDelay` are different? Both of them are commented saying they allow for enough time for the auction + spin up hardware. Are these supposed to be two different phases of time (i.e. I won't know if I need to spin up my hardware until I know if I won the auction)?

Also, there is something a little suspicious about the zero deadline for evicted provers. It does mean that the evicted prover is still responsible for some future publications (up to `exitDelay` from now) but are not given any time where they have monopoly proving rights. Perhaps this is okay because they've already failed to prove so maybe losing the stake based on future publications is just one of the costs.

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -0,0 +1,351 @@
...
+    /// @return deadline The period's deadline timestamp
+    function _closePeriod(Period storage period, uint256 endDelay, uint256 provingWindow)
+        private
+        returns (uint256 end, uint256 deadline)
+    {
+        end = block.timestamp + endDelay;
+        deadline = end + provingWindow;
+        period.end = end;
+        period.deadline = deadline;
+    }
```
LeoPatOZ, line 350:
```suggestion
    function _closePeriod(Period storage period, uint256 _endDelay, uint256 _provingWindow)
        private
        returns (uint256 end, uint256 deadline)
    {
        end = block.timestamp + _endDelay;
        deadline = end + _provingWindow;
        period.end = end;
        period.deadline = deadline;
    }
```

LeoPatOZ: provingWindow is also the name of the global variable

LeoPatOZ:
> Should we unite `livenessWindow` and `provingWindow`?
> 
> * I think we want `provingWindow` to be at least as large as `livenessWindow`, because otherwise a publication can come in on the `end` timestamp, and you will only have `provingWindow` to prove it.
> * On the other hand, there doesn't seem to be a reason for `provingWindow` to be larger than `livenessWindow`.
> 
> Similarly, is there any reason that `exitDelay` and `successionDelay` are different? Both of them are commented saying they allow for enough time for the auction + spin up hardware. Are these supposed to be two different phases of time (i.e. I won't know if I need to spin up my hardware until I know if I won the auction)?
> 
> Also, there is something a little suspicious about the zero deadline for evicted provers. It does mean that the evicted prover is still responsible for some future publications (up to `exitDelay` from now) but are not given any time where they have monopoly proving rights. Perhaps this is okay because they've already failed to prove so maybe losing the stake based on future publications is just one of the costs.

Im really of the opinion that exit and succession delay should be merged - https://github.com/OpenZeppelin/minimal-rollup/pull/49#issuecomment-2714840677

src/protocol/taiko_alethia/ProverManager.sol
ggonzalez94: **TL;DR: since the `evictorIncentive`is calculated from the stake I don't see a case where this would underflow** 

Good question! The only 2 cases where the stake is zero
1. The `livenessBond` parameter was set to zero. 
I don't think this is desirable, because there's no enforcement for a prover to behave honestly. Maybe we should add this check in the constructor? But we are planning to potentially make this an [abstract contract](https://github.com/OpenZeppelin/minimal-rollup/issues/72) and allow overriding some of these values inside virtual functions, so perhaps we can leave the decision until then.
2. `evictProver` is called on a vacant period. This would basically have no effect, since `evictorIncentive` would be zero.

Did I miss any other?

src/protocol/taiko_alethia/ProverManager.sol
ggonzalez94: > In what scenario would period.prover actually be 0 to allow claimProvingVacancy? 

Good question. This is a fallback mechanism that @nikeshnazareth created for what should be a very unlikely scenario. This would only happen if the current prover exits or is evicted, and no one else bids during the `exitDelay` period to become the prover. This in practice will never happen, because even if there are no bidders in the auction you are incentivized to come and set a very high price per publication.

nikeshnazareth (approved):

LeoPatOZ (approved):

PR #52, opened 2025-02-26
Add LibValueTicket and ETHBridge for native token bridging

ernestognw:
Add new ETHBridge.

The design is very simple, and it relies on a new libraries to create and verify what I call "Value Tickets". These are tickets that are signaled and can be proven on the destination chain.

### DIFF
--- a/src/protocol/ETHBridge.sol
+++ b/src/protocol/ETHBridge.sol
@@ -41,10 +41,9 @@ contract ETHBridge is IETHBridge {
     }
 
     function createTicket(uint64 destinationChainId, address to) external payable virtual {
-        address from = msg.sender;
         uint64 blockNumber = block.number.toUint64();
-        ticketId(destinationChainId, blockNumber, from, to, msg.value).signal();
-        emit Ticket(destinationChainId, blockNumber, from, to, msg.value);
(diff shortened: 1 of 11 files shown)


src/protocol/ETHBridge.sol
```diff
@@ -0,0 +1,64 @@
...
+        require(!claimed(id_), AlreadyClaimed());
+        _claimed[id_] = true;
+        _sendETH(ticket.to, ticket.value);
+        emit ETHTicketClaimed(ticket);
+    }
+
+    /// @dev Function to transfer ETH to the receiver but ignoring the returndata.
+    function _sendETH(address to, uint256 value) private returns (bool success) {
+        assembly ("memory-safe") {
+            success := call(gas(), to, value, 0, 0, 0, 0)
```
LeoPatOZ, line 60: I believe @dantaik wanted the ETH bridge to also be able to call an arbitrary contract. (https://github.com/OpenZeppelin/minimal-rollup/pull/61#discussion_r1986649607)
Perhaps we need to rethink the `ValueTicket` struct, as well as changing this call

dantaik: +1

PR #54, opened 2025-02-27
feat: same slot signals

AnshuJalan:


src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -89,4 +98,10 @@ contract TaikoInbox {
             blobIndices[i] = i;
         }
     }
+
+    function _verifySignalSlots(bytes32[] calldata signalSlots) private view {
```
LeoPatOZ, line 102: should we move this to the signal service?

ernestognw (approved):

src/protocol/taiko_alethia/TaikoInbox.sol
ernestognw: We'll have to call the signal service anyway. I agree it may save some gas by avoiding multiple calls, but if there's a function on the signal service that allows batched verification, we still need to copy `signalSlots` to memory, which depending on the array size may carry memory expansion costs that could be bigger. I don't know where the threshold is though

PR #58, opened 2025-02-28
Refactor Signal Service for generalised Message Passing

LeoPatOZ:
Updated:

- merged https://github.com/OpenZeppelin/minimal-rollup/pull/63 changing the signal service to handle both F and S signals.

- Added `sendFastSignal` to signal service (not sure if this is neccessary)
- Bridge no longer handles signalling - all gets delegated to signal service 

--- **Thoughts** ---

Things that are still unclear / open to discussion.

- A lot of `verifyX` functions in the signalling pathways rely on a given state L1 root, however its never verified that this is from a trusted source / the actual L1 state root. I feel like this is an issue. (Also unsure how the commitment syncer fits in all this --> doesn't seem its being used for any signalling pathways)

- Perhaps we could think about unifying some libraries currently we have a lot of layers of indirection e.g the claim ticket flow is as follows: claimTicket --> lib.checkTicket --> lib.verifyTicket --> signal.verifySignal. No strong opinions here however from a DevEX perspective it was harder to implement additional functionality / follow the flow.

### DIFF
--- a/src/libs/LibValueTicket.sol
+++ b/src/libs/LibValueTicket.sol
@@ -37,13 +37,24 @@ library LibValueTicket {
     /// to claim `value` on this chain. It does so by performing an storage proof of `address(this)` on the source chain
     /// using `accountProof` and validating it against the network state `root` using `proof`.
     /// The `root` MUST be trusted.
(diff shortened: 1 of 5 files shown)


AnshuJalan:
LGTM!

PR #61, opened 2025-03-06
Add Signal Service Design Doc

LeoPatOZ:


documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,33 @@
...
+
+In our current protocol, signalling is performed by storing a message at a specific storage slot, then, once the L1 state root is accessible on the L2, executing a storage proof to verify that value exists. This proof can then trigger then certain logic, such as minting tokens.
+
+This is process is described by this diagram.
+
+![Image](https://github.com/user-attachments/assets/5c27c6bc-cdd5-4964-8245-f5ae3f5b0272)
+
+Currently each application is responsible for their own signalling (i.e. there is no central storage of all signals). This means any potential relaying must be done on an app by app basis. We only provide a standard signalling library to store and verify signals on chain.
+
+This approach has latency issues since the latest L1 state root may only update every X number of L2 blocks, causing newer L2 blocks to not have access to the most current L1 state root. This is why there is an alternative method used for message passing.
```
ggonzalez94, line 11: I suggest we provide some details about how and when the L1 state root is updated on the L2. I would also push the "there is an alternative method used for message passing..." to the next section
```suggestion
This approach works well, but has some latency since the latest L1 state root in the case of Taiko is only accessible when the L2 proposer submits a new publication to the L1 and includes a new anchor tx to the L2.
```

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,33 @@
+## Normal ‘slow’ Message Passing
+
+In our current protocol, signalling is performed by storing a message at a specific storage slot, then, once the L1 state root is accessible on the L2, executing a storage proof to verify that value exists. This proof can then trigger then certain logic, such as minting tokens.
+
+This is process is described by this diagram.
+
+![Image](https://github.com/user-attachments/assets/5c27c6bc-cdd5-4964-8245-f5ae3f5b0272)
+
+Currently each application is responsible for their own signalling (i.e. there is no central storage of all signals). This means any potential relaying must be done on an app by app basis. We only provide a standard signalling library to store and verify signals on chain.
```
ggonzalez94, line 9: Should we also describe how the storage slot for signals is calculated and why?

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,33 @@
+## Normal ‘slow’ Message Passing
+
+In our current protocol, signalling is performed by storing a message at a specific storage slot, then, once the L1 state root is accessible on the L2, executing a storage proof to verify that value exists. This proof can then trigger then certain logic, such as minting tokens.
```
ggonzalez94, line 3: Why don't we separate the general message passing from the bridging? I believe the initial section should focus on how general message passing works and then another section can explain bridges as a specific application of message passing

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Future Considerations
+
+- Merging fast and slow in the SS
+- L2 → L1 fast withdrawals using solver [network](https://ethresear.ch/t/fast-and-slow-l2-l1-withdrawals/21161)
+- Currently all signals cannot be cancelled and are stored indefinitely - perhaps we could look into allowing cancellable signals / reusing storage
+
+## Questions
+
+- How do L2 applications have access to ‘trusted’ L1 state root?
```
ggonzalez94, line 31: Isn't this what the anchor transaction provides? Or do you mean something different here?

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,33 @@
...
+- Usefulness of commitment syncer - how is this currently being used to get a trusted root.
+- L2 → L1 messaging
```
ggonzalez94, line 33: Do you mean this still needs to be designed or just documented? Because I believe the design is very similar, just in the other direction

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,33 @@
...
+
+This approach has latency issues since the latest L1 state root may only update every X number of L2 blocks, causing newer L2 blocks to not have access to the most current L1 state root. This is why there is an alternative method used for message passing.
+
+## Same Slot L1 → L2 Message Passing
+
+The signal service (SS) is used for cases where the message is to be sent within the same slot. This [design](https://ethresear.ch/t/same-slot-l1-l2-message-passing/21186) is the one proposed by Nethermind. The basic idea is as follows:
+
+1. The proposer listens to any signals being stored on the L1 SS
+2. The proposer (during his allocated slot) can selectively inject L1 → L2 signals when proposing a batch in `publish` function.
+3. Signals will be verified that they exist in the SS’s storage.
```
ggonzalez94, line 19: Who verifies this? Is the Inbox contract?

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,33 @@
...
+4. A trusted entity (owned by the rollup) will then call `receiveSignal` on the L2 SS. This will fill a mapping of `_receivedSignals` to true (signal —> true)
```
ggonzalez94, line 20: Why do you need a trusted entity? I don't remember this being necessary, nor is discussed in the post

ggonzalez94 (commented):
Looks good overall, but I think:
- We should expand a bit more. While keeping docs concise for now is a good idea, they should still be helpful and easy to read.
- Separate the concepts of general message passing(signals) from bridging
- Make the "fast message passing" section more clear

documentation/SignalServiceDesign.md
LeoPatOZ: From my understanding the rollup operator will need to call 'recieveSignal' on the L2 - see [here](https://github.com/OpenZeppelin/minimal-rollup/pull/45/files#diff-6ae528a96677aaa2ece962de5d7439f828f82e8a6569167c37f4c622db744470R33). This because for 'fast' signals we need to trust that those signals were verified on the L1 since we dont have access to the L1 state root yet. Otherwise anyone could therotically fill the mapping and say the signal was received without providing a proof.

documentation/SignalServiceDesign.md
LeoPatOZ: sorry yes clarified now!

documentation/SignalServiceDesign.md
LeoPatOZ: sorry yes it was meant to be linked to the question below - removed now

ggonzalez94:
@nikeshnazareth @LeoPatOZ @ernestognw  now that I had the time to read again the [research post](https://ethresear.ch/t/same-slot-l1-l2-message-passing/21186) and this doc I believe there might be an under-explored part that we were discussing in person at the hotel lobby.
- "Normal" message passing can be done by using the latest L1 state root available to the L2
- Same slot message passing can be done using the technique described in the post(but it has the downside that it forces the L2 proposer to post to the L1 on the same slot)

Now that we will have preconfs, publishing will be done less often. If I make a deposit from L1->L2 for example I will only be able to claim that once the anchor transaction updates the L1 state root. But if the preconfer can "inject" the signals to the L2 then I can claim my deposit directly on the L2 before the next "sync" happens and then when the proposer submits a publication the Inbox contract can check that those signals are actually present in storage on the L1. This is in practice probably very similar to the mechanism used in same slot message passing, but I think it warrants its own section and thinking.

documentation/SignalServiceDesign.md
LeoPatOZ: okay i re-structured the document!

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,62 @@
...
+
+An example diagram is provided in the ETH bridge section.
+
+---
+
+## Bridging
+
+Bridges are one of the main use cases of signalling.
+
+### ETH Bridge
```
dantaik, line 33: It’s worth noting that the ETH bridge is not just for transferring ETH across chains; it shall be able to interact with arbitrary contracts on the destination chain, passing custom calldata and a specified amount of Ether (msg.value). Essentially, it functions as a remote call bridge, with ETH bridging being the simplest use case

LeoPatOZ: Ah ok i didnt realise the eth bridge would handle contract calls as well - currently it seems to not be able to support this https://github.com/OpenZeppelin/minimal-rollup/blob/bridge/src/protocol/ETHBridge.sol#L60 as we hardcode 0 for the calldata

documentation/SignalServiceDesign.md
```diff
@@ -0,0 +1,62 @@
+## Signal
+
+The Taiko stack refers to cross chain message passing as 'signalling'. The process of signalling consists of writing to a storage location so that it can later be proven on a destination chain using a storage proof.
+
+A signal is stored at a pseudorandom storage slot derived via [erc7201](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ca7a4e39de0860bbaadf95824207886e6de9fa64/contracts/utils/SlotDerivation.sol#L4). This signal consists of the `msg.sender`, `chainId` and `value`.
+
+## Normal ‘slow’ Message Passing
+
+In our current protocol, signalling is performed by storing a message at a specific storage slot, then, once the L1 state root is accessible on the L2, executing a storage proof to verify that value exists. This proof can then trigger then certain logic, such as minting tokens.
```
ggonzalez94, line 9:
```suggestion
In our current protocol, signalling is performed by storing a message at a specific storage slot, then, once the L1 state root is accessible on the L2, executing a storage proof to verify that value exists. This proof can then trigger certain logic, such as minting tokens.
```

ggonzalez94 (approved):
LGTM

documentation/SignalServiceDesign.md
nikeshnazareth: If I'm tracking the code correctly, the "trusted entity" here the TaikoAnchor contract, which is invoked by a hardcoded address that is known publicly. However, the Taiko state-transition-function (enforced by the provers) guarantees that there is exactly one such transaction at the start of every block.

In other words, this is a mechanism to let the proposer specify any signals that they want (which are then checked at publication time). I'll update the description accordingly

nikeshnazareth:
> Now that we will have preconfs, publishing will be done less often. If I make a deposit from L1->L2 for example I will only be able to claim that once the anchor transaction updates the L1 state root. But if the preconfer can "inject" the signals to the L2 then I can claim my deposit directly on the L2 before the next "sync" happens and then when the proposer submits a publication the Inbox contract can check that those signals are actually present in storage on the L1. This is in practice probably very similar to the mechanism used in same slot message passing, but I think it warrants its own section and thinking.

I think this might be implicit in the post, but they're just describing it in language that uses the same mental model that Ernesto has (where L2 transactions are executed when they are posted to L1). Note that the post claims:

> L2 transactions in the batch can... Consume the imported same-slot messages by calling the L2 message contract

They don't make any distinction about where in the batch that happens because it doesn't matter.

They also describe (in section J-2) L2 transactions being able to consume messages at or before the anchor block ID (which can be any block before the publication one), although I think the dependence on the L2 proposer injecting the state should be spelled out a little more, since a user cannot construct such a proof themselves if they don't know which anchor block to use (unless we're talking about an anchor block from a previous publication).

So I think both ideas are implicit in the original post, but I do agree it's worth explaining this in more detail. In particular, in the case where we're discussing L1 messages submitted at or before the anchor block, the relevant proof that the signals were injected correctly, could be executed off-chain.

nikeshnazareth (approved):

documentation/SignalServiceDesign.md
ernestognw: Yes, I'll send a PR improving this section after we merge this.

documentation/SignalServiceDesign.md
ernestognw: I guess @dantaik point is that we should elaborate more on the idea of generalized message passing mechanism. I will open a PR detailing more the general idea, the ETHBridge section can remain simple IMO

ernestognw:
> They don't make any distinction about where in the batch that happens because it doesn't matter.
> 
> They also describe (in section J-2) L2 transactions being able to consume messages at or before the anchor block ID (which can be any block before the publication one), although I think the dependence on the L2 proposer injecting the state should be spelled out a little more, since a user cannot construct such a proof themselves if they don't know which anchor block to use (unless we're talking about an anchor block from a previous publication).

I'd like to expand more on the dependence on the L2 proposer injecting the state root. In reality, it doesn't matter who injects the state root as long as it can be trusted (by whatever mechanism). This opens a design space where a solver network can submit an L1 solution to a full L1->L2->L1 block as [denoted here](https://ethresear.ch/t/same-slot-l1-l2-message-passing/21186#p-51604-tying-a-loop-l1l2l1-6).

Will submit a PR

ernestognw (approved):
Good enough for a first draft

documentation/SignalServiceDesign.md
ernestognw: I think @dantaik has a good point here. Perhaps we want to have a bridge that allows arbitrary calls. If it's isolated (no other functionality), I don't see why that would be an issue.

PR #62, opened 2025-03-06
Prover incentives documentation

ggonzalez94:
Most of the content is the same as the Notion doc, just improved some of the writing and provided more details that were missing in the original doc


documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+- `exitDelay` is the time window between when the prover wants to exit and when the new prover can start their period. It is also the time window between when an inactive prover is ejected and when the new prover can start.
+
+**We call the proving window(the time and the publications that are assigned to a specific prover) a `period`.**
+Provers are only rewarded for the publications they proved within their period and returned their stake after their period is closed(all the publications have been proven).
+
+### Fallback Mechanism
+
+What happens if the prover does not deliver on their promise and does not prove anything after a certain period of time? We now need a new prover to step in. Anyone can calll `eject` to receive a portion of the prover's stake and mark them as inactive. This marks the current prover as ready to be slashed(we will discuss the details in a later section) and allows a new prover to step in(an auction occurs during the `exitDelay` window described in the section above).
```
dantaik, line 50: How about this alternative:

Other provers can also submit proofs at any time. However, if the proof is submitted within the block's period (proving window), the recorded prover remains the auction winner. Otherwise, the recorded prover becomes `msg.sender`.

If the proof used to verify the block is marked as submitted within the period, the proving fee is awarded to the auction winner. Otherwise, the auction winner is slashed for that slot, meaning the stake is tied to each publication.

ggonzalez94: I'm not sure I understand the comment 100% (but I think the confusion is because I need to improve my writing for this section).

> Other provers can also submit proofs at any time. However, if the proof is submitted within the block's period (proving window), the recorded prover remains the auction winner. Otherwise, the recorded prover becomes msg.sender.

Yes, this is the intention. Submitting the proof is actually permisionless(there's no access control on those functions of the contract), but if within their deadline rewards accumulate for the assigned prover, if outisde some the rewards are splitted, and a part of the liveness bond is burnt and the rest goes to the new prover)

>  Otherwise, the auction winner is slashed for that slot, meaning the stake is tied to each publication.
The stake is tied to `Periods` which span multiple publications.

Hopefully the [implementation](https://github.com/OpenZeppelin/minimal-rollup/pull/49) will make some of these details more clear, but let me know if you think there are still big gaps to improve in the doc

ernestognw, line 50:
```suggestion
Whenever a registered prover fails to deliver a proof after a certain time threshold, anyone can call the `eject` function to claim a portion of the inactive prover’s stake and mark them as inactive. This flags the current prover for slashing (details will be discussed later) and initiates an auction during the `exitDelay` window (described in the previous section) to select a new prover.
```

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+
+*In the case no independent prover is willing to step in for whatever reason, or the ones that are collude and decide to charge an insane amount, the system can still maintain liveness assuming the rollup operator is willing to run a prover and is assumed not to want to extract excessive value from their users.*
```
ernestognw, line 52:
```suggestion
*If no independent prover steps in, or if existing provers collude to charge excessively high fees, the system can still maintain liveness. This assumes the rollup operator is willing to run a prover and is expected not to exploit users by extracting excessive value.*
```

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+
+At this point you may wonder what we do with the unproven publications from the offender prover? The way we deal with this currently is by **making proving those publications permisionless**, and rewarding the prover that proves them with a portion of the offender's stake while burning the rest(how much is burned or if it is actually sent to the treasury instead of burned is yet to be defined). The fees for the unproven publications are paid to the new prover, and the fees for the ones that were already proven by the offender are still paid to them.
```
ernestognw, line 54:
```suggestion
You might wonder how we handle unproven publications from the offending prover. Currently, we **make proving these publications permissionless**. The prover who proves them is rewarded with a portion of the offender’s stake, while the remainder is burned (or potentially sent to the treasury—this detail is yet to be finalized). Fees for unproven publications go to the new prover, while fees for publications already proven by the offender still go to them.
```

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Dealing with forced inclusions
+
+**How to deal with forced inclussions is still being discussed. This section describes our current thinking on the subject.**
+
+In our current design, forced transactions are still posted as blobs, but go to a separate queue and are picked when the next proposer calls publish on the Inbox to post their next publication. For now we have decided that they will be posted as a new publication inside the same function call(a new publication is cleaner, easier to price and avoids the risk of a single combined publication being too big to prove).
```
ernestognw, line 60:
```suggestion
In our current design, forced transactions are posted as blobs but are routed to a separate queue. They are picked up when the next proposer calls `publish` on the Inbox to submit their publication. For now, we’ve decided that forced transactions will be posted as a **new publication** within the same function call. This approach is cleaner, easier to price, and avoids the risk of a single combined publication becoming too large to prove.
```

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+
+But we now have a problem to solve: how do we price these delayed publications? Should we charge the forced includer or allow them to free ride the next proposer?
+While we don’t always know exactly the proving cost when their transactions will be included(since the current prover turn might be over by the time this publication is included), allowing a free ride creates negative incentives, where actors are incentivized to batch as many transactions as they can and post them in a delayed publication, giving them a lower execution price as long as they are willing to wait that time for their inclusion. This penalizes the prover, who now has to prove a very big publication that he is not compensated for(or the proposer if we make him pay for the publication).
```
ernestognw, line 63:
```suggestion
However, this raises a question: **How should we price these delayed publications?** Should the forced includer pay, or should they free-ride on the next proposer? 

While the exact proving cost at the time of inclusion may be uncertain (since the current prover’s term might have ended), allowing free-riding creates negative incentives. Actors could batch as many transactions as possible into delayed publications, securing a lower execution price by waiting for inclusion. This penalizes the prover (who must prove a large publication without adequate compensation) or the proposer (if they are forced to cover the cost).
```

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+- This also has the downside that you have to store multiple checkpoints in the Inbox contract, because proving may happen out of order.
+
+## Proposed design
+
+Anyone that wants to be a prover can register as the designated prover by signaling how much they are willing to charge **on a per publication basis**. The prover that offers the lower amount gets monopoly rights to prove publications until someone else comes in and offers a lower cost per publication. This way proposers know exactly how much they will have to pay to get their publication proven, and are not required to put additional capital, making the system more capital efficient.  
+New participants that want to undercut the current prover need to offer a lower price per publication that **is at least a defined percentage lower** than the current prover or other participants. This is to avoid gas wars common in auctions where the new participant just offers one wei less than the current best offer.  
+
```
ernestognw, line 26: We're missing a section here that details better how the undercutting percentage is agreed or calculated. I was thinking 0.1 ETH as (iirc) more than 99% of the L2 transactions are worth less than that.

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+The proposed solution is to ask the delayed proposer to also pay a proving fee, **but this should be higher than the regular proving fee** because of:
+
+1. Their publication are potentially much larger than the regular proposer publication(since they can accumulate transactions for an arbitrarily large amount of time and are limited by DA only)
+2. Proving cost is not always known, so it might actually go up by the time it is time to prove their publication
+3. We want to disincentivize the formation of a market where an actor batches multiples transactions from people and include that with a delay just because it is cheaper. Unless you are getting censored you should go trough the regular flow.
```
ernestognw, line 69:
```suggestion
The proposed solution is to require the delayed proposer to pay a proving fee, **set higher than the regular fee** for the following reasons:

1. **Size:** Delayed publications can accumulate transactions over an arbitrarily long time, making them potentially much larger than regular publications (limited only by data availability).
2. **Uncertainty:** Proving costs may increase by the time the publication is included.
3. **Incentives:** We want to discourage the formation of a market where actors batch transactions for delayed inclusion simply because it’s cheaper. Unless censored, users should follow the regular flow.
```

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+
+- After the pacaya fork, there is now an option to prove multiple non consecutive batches(and anyone can do it), but if the batch is within its proving window(60 minutes as of this writing) the rewards goes to the proposer.
+This is a bit more flexible, but the incentives are basically the same - each proposer is incentivized to prove or get a prover for their batches only, which does not result in the most efficient proving costs. If the proposers are sophisticated enough, an off-chain market may develop, where different proposers pay the same prover to generate a proof for their batch, and the prover may aggregate those proofs in order to get a more efficient result. This is still uncertain, and it may even have a very centralizing effect, since provers are incentivized to make a deal with as many proposers as possible to ensure they can prove consecutive batches and get paid off-chain(if different batches have proposers that have chosen different provers then none of them will be incentivized to prove the entire set).
+- This also has the downside that you have to store multiple checkpoints in the Inbox contract, because proving may happen out of order.
```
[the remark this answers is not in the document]
ggonzalez94: I don't think threre's a significant difference between the two? But I do think including "downside" makes it more clear, so keeping the original version for now

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+
+Anyone that wants to be a prover can register as the designated prover by signaling how much they are willing to charge **on a per publication basis**. The prover that offers the lower amount gets monopoly rights to prove publications until someone else comes in and offers a lower cost per publication. This way proposers know exactly how much they will have to pay to get their publication proven, and are not required to put additional capital, making the system more capital efficient.  
+New participants that want to undercut the current prover need to offer a lower price per publication that **is at least a defined percentage lower** than the current prover or other participants. This is to avoid gas wars common in auctions where the new participant just offers one wei less than the current best offer.  
+
+*This system effectively behaves as a reserve english ahead of time auction.*
+
+![Prover auction](./images/prover-market.png)
+
+While publications will differ in proving cycles they require, the prover should account for this variation and since we don’t have a proper way to calculate this on the L1 they should choose this value on an average case basis.
```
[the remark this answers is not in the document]
ggonzalez94: Agree this is more readable. I'm only changing it slightly to:
```
While publications vary in the proving cycles they require, provers should account for this variability. Since L1 lacks a way to calculate these differences, provers should base their bids on an average case.
```

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+- This also has the downside that you have to store multiple checkpoints in the Inbox contract, because proving may happen out of order.
+
+## Proposed design
+
+Anyone that wants to be a prover can register as the designated prover by signaling how much they are willing to charge **on a per publication basis**. The prover that offers the lower amount gets monopoly rights to prove publications until someone else comes in and offers a lower cost per publication. This way proposers know exactly how much they will have to pay to get their publication proven, and are not required to put additional capital, making the system more capital efficient.  
+New participants that want to undercut the current prover need to offer a lower price per publication that **is at least a defined percentage lower** than the current prover or other participants. This is to avoid gas wars common in auctions where the new participant just offers one wei less than the current best offer.
```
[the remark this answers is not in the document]
ggonzalez94: You are correct, I changed the wording slightly

documentation/prover-incentives.md
```diff
@@ -0,0 +1,69 @@
...
+### Transition Periods
+
+The ideal system is one where a prover has the right for an extended period of time. This allows them to aggregate proofs for multiple publications and post them in a single L1 transaction, reducing the on-chain costs.  
+We want to avoid scenarios where provers are constantly changing, since this would result in less efficient proving costs and a lot of unpredictablity for provers. The good news is that this should happen naturally, as the price per publication they charge should converge to the cost of proving. But we still introduce a transition period when a prover is outbid or wants to exit.
+
+- `succesionDelay` is the time window between when the prover is outbid and when the new prover can start their period. During this time other interested provers or the existing one can come in and offer an even lower price.
+- `exitDelay` is the time window between when the prover wants to exit and when the new prover can start their period. It is also the time window between when an inactive prover is ejected and when the new prover can start.
+
+**We call the proving window(the time and the publications that are assigned to a specific prover) a `period`.**
```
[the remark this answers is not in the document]
ggonzalez94: It is a core abstraction in the implementation, so I would keep it. I'll review the doc again since maybe it should be used more extensively here

documentation/prover-incentives.md
ernestognw: Ah right, I got messed up between two concepts. On one side there's the need of documenting the mechanics of the percentage work, and on the other we need a default fee value for each `period`. I see there's an [`initialFee` in the `ProverManager` constructor](https://github.com/OpenZeppelin/minimal-rollup/pull/49/files#diff-d43081579a98c4c833bda7e2444609810a56b9fb193513d8f66eed2232a113a5R114).

It's true the design document may not specify any implementation details, so I'm reconsidering whether we should document it here.

For both things, I the framework to get the minimal-rollup up and running should have defaults, and we will document tradeoffs.

ernestognw (approved):
Looking good!

documentation/prover-incentives.md
```diff
@@ -0,0 +1,72 @@
...
+- Providing proposers with predictable costs when submitting publications.
+- Simple design and minimal L1 gas costs.
+
+*The specifics of the mechanism are still being discussed, and some of the details might change as we do more research on the incentives.*
+
+## The problem
+
+Verifying proofs on Ethereum L1 is a [gas-intensive operation](https://docs.alignedlayer.com/#why-are-we-building-aligned). For example, Groth16 proofs, among the cheapest, cost around 250k gas, while STARKS can exceed 1M gas. As a result, most zk-rollups call their inbox contract's verification function only sporadically.
+
+For based validity rollups, achieving a similar property raises the question: Who should verify batches submitted to L1, and how can proposers be incentivized to submit valid batches if proofs aren’t generated immediately? A common solution is to require proposers to deposit a bond when submitting a batch. If they fail to provide a validity proof within a set time window, they lose their stake, and proving becomes permissionless. This approach is currently used by [Taiko].
```
nikeshnazareth, line 16: "[Taiko]". Is this supposed to be a link?

nikeshnazareth (approved):

documentation/prover-incentives.md
ernestognw: Yes, [this](https://docs.taiko.xyz/resources/terminology/#validity-bond) should be it

PR #64, opened 2025-03-06
Describe proposed blob structure

nikeshnazareth:


documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+
+The rollup node should find relevant publication hashes in the `PublicationFeed` contract. This will include anything published by its inbox plus any other data feeds that it is watching. There may be multiple relevant publications that share the same blobs, and the rollup node will need to track which blobs it is processing for which reasons. In the simplest case it is just looking for its own rollup transactions inside publications made by its own inbox.
+
+### Publication Headers
+
+For each relevant hash, the publication is the concatenation of all the corresponding blobs. The rollup node should retrieve and validate the first 31-byte chunk, and interpret it as a header with the structure:
+
+- the version (set to zero) (1 byte)
+- header length in chunks, excluding this one (1 byte)
+    - I expect this to be zero most of the time
```
dantaik, line 72: Why header length is zero most of the time?

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+- Up to 5 data type records consisting of
+    - data identifier (3 bytes)
+    - offset to corresponding segment after accounting for the multiplier (2 bytes)
+- Pad the remaining space with zeroes. Note that zero is an invalid data identifier, so there should be no confusion about what counts as padding.
+
+If the header is larger than one chunk, retrieve and validate the remaining chunks. They should all be structured with the same data type records (up to 6 in a chunk) and padded with zeros.
+
+Although we are not using binary search, I still think it is useful to have a deterministic structure. Therefore, in addition to validating the header structure, the rollup nodes should ensure:
+
+- all data identifiers are listed in ascending order
```
dantaik, line 85: If we do not use binary search, will the order of data identifiers matter?

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+    - note that the size of a segment is implicitly defined by where the next segment starts.
+
+Assuming the publication header is valid, the rollup node will know which chunks to retrieve and validate to find its relevant segments.
+
+
+### Compressed segments
+
+The data identifier will indicate which kind of compression (eg. gzip, lz4, etc) was used. After decompression, the segment should use the same structure so that it can be further subdivided. For simplicity:
+
+- the segment header should be outside the compression. This means the expected contents can be identified (and potentially ignored) before retrieving the whole segment and decompressing it.
```
dantaik, line 97: there is no definition of "segment header"...

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+- we should disallow nested compressed segments. I suspect they are unlikely to achieve much (you can always define the data identifier as multiple rounds of a compression algorithm) and we don't want the node to have to recurse indefinitely only to find the publication is poorly formed.
+
+### Rollup segments
```
dantaik, line 100: If a rollup consists of multiple segments, should it concatenate these segments (excluding their headers) into a single byte array and validate them as a whole, or should each segment be validated individually, with the results (potentially structured) then concatenated together?

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+    - data identifier (3 bytes)
+    - offset to corresponding segment after accounting for the multiplier (2 bytes)
+- Pad the remaining space with zeroes. Note that zero is an invalid data identifier, so there should be no confusion about what counts as padding.
+
+If the header is larger than one chunk, retrieve and validate the remaining chunks. They should all be structured with the same data type records (up to 6 in a chunk) and padded with zeros.
+
+Although we are not using binary search, I still think it is useful to have a deterministic structure. Therefore, in addition to validating the header structure, the rollup nodes should ensure:
+
+- all data identifiers are listed in ascending order
+- the actual data segments are also in the same order (i.e. the segment offsets are strictly increasing)
```
dantaik, line 86: I prefer not to use the next segment’s offset to determine the current segment’s size. This way, app A can use segment [100, 200], while app B can use segment [100, 300], allowing for greater flexibility.

documentation/Blob Structure.md
nikeshnazareth: because I suspect the 5 segments already described by the first chunk will be sufficient. We only need one segment per content type (eg. rollup, or compression mechanism). I expanded on this comment in the doc

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+    - I expect this to be zero most of the time
+- multiplier (1 byte)
+    - the log base 2 of a scaling factor for data offsets
+    - i.e. if this value is $x$ and an offset is specified as $y$, the relevant data starts at chunk $2^x \cdot y$
+- Up to 5 data type records consisting of
+    - data identifier (3 bytes)
+    - offset to corresponding segment after accounting for the multiplier (2 bytes)
+- Pad the remaining space with zeroes. Note that zero is an invalid data identifier, so there should be no confusion about what counts as padding.
+
+If the header is larger than one chunk, retrieve and validate the remaining chunks. They should all be structured with the same data type records (up to 6 in a chunk) and padded with zeros.
```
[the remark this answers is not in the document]
nikeshnazareth: Yes, I think it makes sense to think of the whole structure as corresponding to one file. The header is the start of the file, which spans as many blobs as necessary.

documentation/Blob Structure.md
nikeshnazareth: I don't think the order matters, but my intuition is that data formats should be as deterministic/structured as possible by default. I think things like json are intentionally flexible because they're partially intended to be human readable.

If we insist on ascending order then:
- two publications can be checked for equality directly (by comparing their bytes or their hashes)
- we can reintroduce binary search if the sizes ever get large enough

This is a very mild preference because I don't see a particular reason to prefer one over the other.

documentation/Blob Structure.md
nikeshnazareth: I included a comment emphasising that it's the same format as the publication header

documentation/Blob Structure.md
nikeshnazareth: I'm not sure what we're validating here but I think you wouldn't concatenate them in either case. Any data that should be processed together should be part of the same segment. If a rollup processes multiple segments, it will be because they're broken down into semantically different things like:
- the rollup transactions
- another rollup's transactions
- coordination information with other rollups
- potentially configuration information or updates that do not require releasing new node software

So the idea would be:
- use the headers to identify and locate the relevant segments
- if there are compressed segments, use the headers to ensure they are relevant and then decompress / locate the relevant segments
- you now have different segments that need to be processed independently

documentation/Blob Structure.md
nikeshnazareth: I think we're conceptualising "segments" differently, and perhaps that's relevant to [this comment](https://github.com/OpenZeppelin/minimal-rollup/pull/64#discussion_r1988377589), whether it matters that the segments are ordered, and whether we need a length field.

I imagine each segment contains a particular type of information (where the type is identified by the data format id). In your example, something about the data needs to change at position 200 for app A to decide to stop processing it (eg. perhaps 100-200 represents A's transactions, while B embeds rollup A so it needs to know about A's transactions, but also needs to care about its own transactions in 200-300). In that case, I'd still call 100-200 one segment, and 200-300 another segment, and B would have registered that it wants to process both types.

Can you think of a scenario where rollups process partially overlapping information, but it does not make sense to conceptually divide the overlapping part as its own data segment?

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+    - the header is expected to be sorted by application ID, and each application will look for their record (which tells them where to find the actual data) using a binary search
+    - applications must use this search mechanism precisely. If the header is not sorted correctly or the binary search fails for whatever reason, the data should be considered missing (or junk). We do not want a scenario where different clients reach different conclusions about what data is stored for a given application ID
+    - even though each of our publications can span several blobs, I expect there will only be a handful of rollups per publication (and they may even be grouped under the same compression id). Instead of doing a binary search, I recommend applications do a linear search through the whole header to ensure it has the expected structure (ie. sorted application IDs)
+
+
+## The proposal
+
+With these modifications to Dankrad's proposal, here is a complete description of my suggestion. I will describe what an L2 rollup node shoud do to update it state. The corresponding protocol for proposers and provers should follow pretty naturally.
```
ernestognw, line 51:
```suggestion
With these modifications to Dankrad's proposal, here is a complete description of the suggested design. The description is for how an L2 rollup node must update its state. The corresponding protocol for proposers and provers should follow naturally.
```

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+
+The rollup node should find relevant publication hashes in the `PublicationFeed` contract. This will include anything published by its inbox plus any other data feeds that it is watching. There may be multiple relevant publications that share the same blobs, and the rollup node will need to track which blobs it is processing for which reasons. In the simplest case it is just looking for its own rollup transactions inside publications made by its own inbox.
+
+### Publication Headers
+
+For each relevant hash, the publication is the concatenation of all the corresponding blobs. The rollup node should retrieve and validate the first 31-byte chunk, and interpret it as a header with the structure:
+
+- the version (set to zero) (1 byte)
+- header length in chunks, excluding this one (1 byte)
+    - I expect this to be zero most of the time, because this chunk can identify 5 segments, which I suspect is sufficient for most publications.
```
ernestognw, line 72:
```suggestion
    - Most of the times, this field is expected to be `0` since this chunk accommodates 5 segments already; likely enough for most publications.
```

documentation/Blob Structure.md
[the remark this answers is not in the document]
ernestognw: Yes, so the header spans only at the beginning of the first blob since we're not expecting a significant amount of `application_id`. In theory, a header can even span more than 1 blob but I don't think that's realistic

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+    - the log base 2 of a scaling factor for data offsets
+    - i.e. if this value is $x$ and an offset is specified as $y$, the relevant data starts at chunk $2^x \cdot y$
+- Up to 5 data type records consisting of
+    - data identifier (3 bytes)
+    - offset to corresponding segment after accounting for the multiplier (2 bytes)
+- Pad the remaining space with zeroes. Note that zero is an invalid data identifier, so there should be no confusion about what counts as padding.
+
+If the header is larger than one chunk, retrieve and validate the remaining chunks. They should all be structured with the same data type records (up to 6 in a chunk) and padded with zeros.
+
+Although we are not using binary search, I still think it is useful to have a deterministic structure. Therefore, in addition to validating the header structure, the rollup nodes should ensure:
```
ernestognw, line 83:
```suggestion
Although the proposal doesn't rely on binary search, it is still useful to have a deterministic structure. Therefore, in addition to validating the header structure, the rollup nodes should ensure:
```

documentation/Blob Structure.md
ernestognw: I would support the ascending order approach, I would expect linear search to be simpler for small lists. Here we expect around 5 segments (or `application_id`s).

> I don't think the order matters, but my intuition is that data formats should be as deterministic/structured as possible by default.

Does a structured data format also helps to identify invalid chunks? I also feel checking the order is not needed as long as the application knows how to read their own data

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+
+- all data identifiers are listed in ascending order
+- the actual data segments are also in the same order (i.e. the segment offsets are strictly increasing)
+- all segment offsets point within the publication
+    - note that the size of a segment is implicitly defined by where the next segment starts.
+
+Assuming the publication header is valid, the rollup node will know which chunks to retrieve and validate to find its relevant segments.
+
```
ernestognw, line 91:
```suggestion
```

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Compressed segments
+
+The data identifier will indicate which kind of compression (eg. gzip, lz4, etc) was used. After decompression, the segment should use the same structure so that it can be further subdivided. For simplicity:
+
+- the segment header (which has the same format as the publication header) should be outside the compression. This means the expected contents can be identified (and potentially ignored) before retrieving the whole segment and decompressing it.
+- we should disallow nested compressed segments. I suspect they are unlikely to achieve much (you can always define the data identifier as multiple rounds of a compression algorithm) and we don't want the node to have to recurse indefinitely only to find the publication is poorly formed.
```
ernestognw, line 98: I feel we want to discourage nested compressed arguments rather than disallowing them. The reasoning is that I don't see how to enforce disallowing them.

```suggestion
- Nested compressed segments must be discouraged. The compression rounds can be identified by a _format identifier_ (i.e. `application_id`) and it's not preferrable for the node to recurse indefinitely only to find a poorly-formed publication.
```

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Rollup segments
+
+After finding and decompressing all relevant segments, the rollup node should process them. The data structure should be defined by the particular use case, but I would recommend
```
ernestognw, line 102:
```suggestion
After finding and decompressing all relevant segments, the rollup node should process them. The data structure should be defined by the particular use case, with the following recommendations:
```

ernestognw (commented):
Looking good. I reworded some sections, mainly aiming for clarity and to make the writing less personal

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
+# Blob Structure
+
+## Context
+
+In the [Overall Design document](./Overall%20Design.md) we noted that blobs can be shared between multiple rollups. Here we propose a possible supporting blob structure.
+
+The design goals are:
```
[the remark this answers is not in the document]
nikeshnazareth:
```suggestion
Blobs can be shared between multiple rollups, as noted in the [Overall Design document](./Overall%20Design.md). The proposed blob structure is guided by the following design goals:
```

Thanks. Just using "proposed blob structure" for clarity

documentation/Blob Structure.md
nikeshnazareth:
```suggestion
- Nested compressed segments are discouraged to avoid nodes recursing indefinitely only to find a poorly-formed publication. However, format identifiers are generic enough to represent any kind of data, including the number of compression rounds.
```

Fair enough. Just rephrasing for style.

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+- if two rollups share a compression algorithm, both transaction lists will need to be decompressed together (if they want to take advantage of shared compression), so in that case they should also be a single leaf, and we will need an additional mechanism to describe the split within the compressed blob.
+
+It seems a flat structure (which is one of the options they mention) is both simpler and more efficient.
+
+Their proposal also requires registering with the blob aggregation service so it knows which DA layer and compression mechanism to use. Moreover, the rollups are responsible for ensuring the blobs they retrieve out of the aggregation service match the ones that were passed in (using a signature).
+
+ In our case we are only interested in Ethereum as the DA layer, and we would like to avoid introducing an additional off-chain service. Since the supported compression algorithms are unlikely to change often, we can let rollups register them on chain so individual L1 proposers can see the registrations.
```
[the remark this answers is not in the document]
nikeshnazareth:
```suggestion
The minimal rollup scope only considers Ethereum as the DA layer, avoiding introducing an additional off-chain service. Since the supported compression algorithms are unlikely to change often, rollups may register them on chain so individual L1 proposers can choose their preferred algorithm.
```

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Dankrad's proposal
+
+Dankrad has create [this proposal](https://ethresear.ch/t/suggested-format-for-shard-blob-header/9996). This seems suitable for our purposes, except I would make the modifications:
+
+- each supported compression algorithm could map to one of the "application ids" (so it's more of a data format identifier rather than an application identifier)
+    - in this way, multiple rollups can find and decompress the same compressed data. We will then need a way to split the decompressed data by rollups. I suggest using the same format inside the compressed segment.
```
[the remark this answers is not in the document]
nikeshnazareth:
```suggestion
- identify each supported compression algorithm with an `application_id` as if it were a _format_ identifier
    - Multiple rollups can find and decompress the same compressed data, and a way to split the decompressed data will be required. The suggested approach is using the same segmenting scheme inside the compressed segment
```

What I mean is something like:
- Rollup A transactions have some identifier (let's call it `A`)
- Rollup B transactions have some identifier (let's call it `B`)
- Both rollups support Gzip, which has its own identifier (let's call it `G`)

I think we should compress all the transactions together (for better compression ratios) and then describe this as one Gzip segment. So both rollups recognise they support `G` and decided to process it. After decompressing the data we will have one segment that contains two kinds of transactions (those destined for rollup A, and those destined for rollup B), and we need a way for each rollup to identify its own transactions. I'm suggesting we use the same data format (i.e. introduce a header declaring two segments `A` and `B` pointing to the actual data). As noted below, I think this header should be outside the compression (but still part of `G` segment's data) so rollup C can immediately discard it without doing the decompression.

I will change this to "using the same segmenting scheme" to make it clearer

ernestognw (approved):
LGTM

documentation/Blob Structure.md
```diff
@@ -0,0 +1,106 @@
...
+- his post allows for up to 1535 applications per blob. This has some cascading effects:
+    - the header is expected to be sorted by application ID, and each application will look for their record (which tells them where to find the actual data) using a binary search
+    - applications must use this search mechanism precisely. If the header is not sorted correctly or the binary search fails for whatever reason, the data should be considered missing (or junk). We do not want a scenario where different clients reach different conclusions about what data is stored for a given application ID
+    - even though each of our publications can span several blobs, I expect there will only be a handful of rollups per publication (and they may even be grouped under the same compression id). Instead of doing a binary search, I recommend applications do a linear search through the whole header to ensure it has the expected structure (ie. sorted application IDs)
```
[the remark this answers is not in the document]
nikeshnazareth:
```suggestion
    - even though publications can span several blobs, it's expected that only a handful of rollups will publish together, perhaps grouped under the same _compression id_ (i.e. `application_id`). Instead of a binary search, a linear search through the whole header can be used to ensure ids are sorted.
```

It's not strictly required, so I changed the phrasing to "can be used"

PR #70, opened 2025-03-11
Simplify CheckpointTracker

### LINKED ISSUE #56: Simplify the CheckpointTracker by storing only the latest checkpoint
Today we are allowing for parallel proving and because of that we need to store multiple transitions. If we implement #15 on its current proposed version we can rely on sequential proving and potentially we don't need to store other transitions.

ggonzalez94:
Fixes #56 

We discussed that with the correct incentive mechanism for provers(#49 ) we could get rid of parallel proving and the concept of transitions all together. This PR changes the CheckpointTracker to store only the latest checkpoint. Older checkpoints should be synced to the SignalService or CheckpointSyncer once we decide where we want to store them.

- Only store the latest checkpoint hash
- Modify unit tests

### DIFF
--- a/src/protocol/CheckpointTracker.sol
+++ b/src/protocol/CheckpointTracker.sol
@@ -12,19 +12,11 @@ contract CheckpointTracker is ICheckpointTracker {
     /// the state of the rollup at a specific point in time
     bytes32 public provenHash;
 
-    /// @notice Verified transitions between two checkpoints
-    /// @dev the start checkpoint is not necessarily valid, but the end checkpoint is correctly built on top of it.
-    mapping(bytes32 startHash => bytes32 endHash) public transitions;
-
     IPublicationFeed public immutable publicationFeed;
 
     // This would usually be retrieved dynamically as in the current Taiko implementation, but for simplicity we are
     // just setting it in the constructor
     IVerifier public immutable verifier;
-
-    /// @notice The maximum number of additional checkpoint transitions to apply in a single proof
-    /// @dev This limits the overhead required to submit a proof
-    uint256 private constant MAX_EXTRA_UPDATES = 10; // TODO: What is a reasonable number here?
 
     /// @param _genesis the checkpoint commitment describing the initial state of the rollup
     /// @param _publicationFeed the input data source that updates the state of this rollup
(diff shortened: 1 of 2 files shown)


src/protocol/CheckpointTracker.sol
```diff
@@ -44,52 +36,19 @@ contract CheckpointTracker is ICheckpointTracker {
...
-        require(start.publicationId < end.publicationId, "Start must be before end");
+        bytes32 startCheckpointHash = keccak256(abi.encode(start));
+        require(startCheckpointHash == provenHash, "Start checkpoint must be the latest proven checkpoint");
 
+        require(start.publicationId < end.publicationId, "End publication must be after the last proven publication");
+
+        bytes32 startPublicationHash = publicationFeed.getPublicationHash(start.publicationId);
         bytes32 endPublicationHash = publicationFeed.getPublicationHash(end.publicationId);
-        require(endPublicationHash != 0, "Publication does not exist");
+        require(endPublicationHash != 0, "End publication does not exist");
```
ggonzalez94, line 47: I'm not sure we need this check. The way the `PublicationFeed` is implemented it uses an array for checkpoints. In the case a checkpoint does not exist the function `getPublicationHash` itself will revert because the index is out of bounds.
But if we want to remain agnostic of the implementation and rely solely on the interface then it does make sense to keep the check

src/protocol/CheckpointTracker.sol
```diff
@@ -44,52 +36,19 @@ contract CheckpointTracker is ICheckpointTracker {
...
-        );
+        verifier.verifyProof(startPublicationHash, endPublicationHash, start.commitment, end.commitment, proof);
 
+        provenHash = keccak256(abi.encode(end));
         emit TransitionProven(start, end);
-
-        bytes32 endHash = keccak256(abi.encode(end));
-        if (startHash == provenHash) {
-            provenHash = endHash;
-            emit CheckpointUpdated(endHash);
```
LeoPatOZ, line 70: minor point: we are no longer emitting the `CheckpointUpdated` event (before we emitted both `TransitionProven` and `CheckpointUpdated`). Not sure if this is needed or not.

ggonzalez94: We only need one event now, since we no longer have proofs that don't update the checkpoints and proofs that do - now all of them advance the chain.
(I'm open to choosing a better name or changing the event parameters though)

src/protocol/CheckpointTracker.sol
dantaik: I think we can remove this check if we added a net-spec comment in getPublicationHash about the revert.

dantaik (approved):

PR #73, opened 2025-03-13
Improve SignalService docs

ernestognw:
Extends the documentation of the signal service. In particular, it addresses some of the comments in the past PR.

### Generalization of signaling

The new version tries to expand on the idea of signaling being general enough. For example, one can build their own app to "fast-track" their own trusted values. This way, the mechanics of doing so are passed to the application developers (i.e. Solidity developers) so that they can incentivize the execution whichever they want.

### Preconfirmations

By generalizing the signaling mechanism, it's possible to treat signals differently depending on the context. In the context of preconfirmations, a rollup may decide to condition the inclusion of "fast-tracked" signals in their inbox. I see this as a feature that should be customizable when using the library (e.g. `contract TaikoInbox is BaseInbox, EnforcedFastSignalInbox { ... }`)


dantaik (approved):

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
-## Signal
+# Signals
 
-The Taiko stack refers to cross chain message passing as 'signalling'. The process of signalling consists of writing to a storage location so that it can later be proven on a destination chain using a storage proof.
+The minimal rollup stack refers to cross-chain message passing as "signalling". Signalling involves writing to a storage location so that it can later be proven on a **destination chain** using a storage proof for a trusted L1 state root.
 
-A signal is stored at a pseudorandom storage slot derived via [erc7201](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ca7a4e39de0860bbaadf95824207886e6de9fa64/contracts/utils/SlotDerivation.sol#L4). This signal consists of the `msg.sender`, `chainId` and `value`.
+A storage proof is a [Merkle proof](https://en.wikipedia.org/wiki/Merkle_tree) showing that a value is stored at a specific storage slot. This proof is generated by the L1 chain and can be verified by the L2 chain as long as the root hash of the **source chain** state is available on the **destination chain**.
```
LeoPatOZ, line 5:
```suggestion
A storage proof is a [Merkle proof](https://en.wikipedia.org/wiki/Merkle_tree) showing that a value is stored at a specific storage slot. This proof is generated by the **source chain** and can be verified by the **destination chain** as long as the root hash of the **source chain** state is available on the **destination chain**.
```
should we make this more general and just stick with source and dest chain? As its not always L1 -> L2?

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
-## Signal
+# Signals
 
-The Taiko stack refers to cross chain message passing as 'signalling'. The process of signalling consists of writing to a storage location so that it can later be proven on a destination chain using a storage proof.
+The minimal rollup stack refers to cross-chain message passing as "signalling". Signalling involves writing to a storage location so that it can later be proven on a **destination chain** using a storage proof for a trusted L1 state root.
```
LeoPatOZ, line 3:
```suggestion
The minimal rollup stack refers to cross-chain message passing as "signalling". Signalling involves writing to a storage location so that it can later be proven on a **destination chain** using a storage proof for a trusted **source chain** state root.
```
same idea with being generic as which state root we are referring to

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+The simplest form of signalling requires waiting for the next **source chain** state root to be available on the **destination chain** via a trusted mechanism. In the case of L1 → L2 communication, the latest available L1 root is included in the L2 batch by the proposer in an anchor transaction.
 
-An example diagram is provided in the ETH bridge section.
+Using the signal requires the following steps:
+
+1. Store a signal on the L1.
+2. Wait for the next L2 batch to be proposed on the L1.
+3. Verify the signal in your L2 application using the latest L1 root to trigger some logic.
+
+However, the waiting period forbids immediate communication between the two chains since users need to wait for the next L1. This is where the 'fast' signalling pathway comes in.
```
LeoPatOZ, line 29:
```suggestion
However, the waiting period forbids immediate communication between the two chains since users need to wait for the next L1 state root to be available. This is where the 'fast' signalling pathway comes in.
```

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+
+#### Same Slot L1 → L2 ('fast') Signalling
+
+In the 'fast' signalling pathway, the L2 proposer can assist an application to bypass the waiting period for the next L1 state root by injecting the signal into the L2 chain. This is done by the L2 proposer listening to the L1 signal and including it in the L2 block proposal.
+
+Using the signal requires the following steps:
+
+1. An application on the L1 stores a signal in the SS.
+2. The L2 proposer listens to any signal being stored on the L1 SS.
```
LeoPatOZ, line 38:
```suggestion
2. The L2 proposer listens to any signals being stored on the L1 SS.
```

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+
+However, the waiting period forbids immediate communication between the two chains since users need to wait for the next L1. This is where the 'fast' signalling pathway comes in.
+
+#### Same Slot L1 → L2 ('fast') Signalling
+
+In the 'fast' signalling pathway, the L2 proposer can assist an application to bypass the waiting period for the next L1 state root by injecting the signal into the L2 chain. This is done by the L2 proposer listening to the L1 signal and including it in the L2 block proposal.
+
+Using the signal requires the following steps:
+
+1. An application on the L1 stores a signal in the SS.
```
LeoPatOZ, line 37:
```suggestion
1. An application on the L1 stores a signal in the Signal Service (SS).
```

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+2. The L2 proposer listens to any signal being stored on the L1 SS.
+3. After building a block and obtaining a commitment, the L2 proposer submits a transaction batch to their L1 inbox contract. This batch includes an anchor transaction where the proposer injects the L1 state root and the signals that were stored on the L1 SS.
+4. The inbox contract verifies that the signals exist in the L1 SS using a storage proof with the L1 state root that was just injected.
```
LeoPatOZ, line 40: Im confused by step 3/4.
```
The inbox contract verifies that the signals exist in the L1 SS using a storage proof with the L1 state root that was just injected.
```
However the inbox contract doesnt rely on an 'injected L1 state root' to verify signals as the inbox is on the L1. It simply reads the signal services storage directly.

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+Considering that [_relying on same-slot L1 messages for L2 execution creates a tight coupling between L1 and L2_](http://ethresear.ch/t/same-slot-l1-l2-message-passing/21186#p-51604-are-shared-sequencersbuilders-needed-here-7) as they would _reorg_ together, it's recommended for applications developers to build mechanisms that are resilient to reorgs. However, the ability to create atomic L1 <> L2 execution remains, as they can always write applications that enforce atomicity (e.g. [fast (and slow) L2 -> L1 withdrawals](https://ethresear.ch/t/fast-and-slow-l2-l1-withdrawals/21161)).
 
 ---
 
-## Bridging
+## Application Examples
+
+### State root fast-tracking
+
+An application that requires immediate access to the latest L1 state root can use the fast signalling to fast-track it into the L2 in a permissioned way. The mechanics to deal with invalid or untrusted state roots are left to the application.
```
LeoPatOZ, line 58:
```suggestion
An application that requires immediate access to the latest L1 state root can use the fast signalling to fast-track it into the L2 in a permissioned way. The mechanisms to deal with invalid or untrusted state roots are left to the application.
```

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+4. The inbox contract verifies that the signals exist in the L1 SS using a storage proof with the L1 state root that was just injected.
+5. The L2 proposer calls `receiveSignal` on the L2 SS through the anchor transaction executed at the start of L2 blocks. This fills a mapping of `_receivedSignals` to true (signal —> true).
+6. Any application that requires same slot signalling will just verify the signal is found in the `_receivedSignals` mapping (i.e. they don’t need to provide a storage proof).
+
+![Fast Signaling](./fast_signaling.svg)
+
+A corner case is that where an L2 proposer decides not to receive one of the signals stored in the L1 SS within the anchor transaction, making any application using the signal (as in step 6) revert. In those cases, the application can always fallback to the 'slow' signalling pathway given the signal can't be deleted and the user can always wait for the next L1 state root. Applications can build mechanisms to incentivize proposers to include their signals in the L2 anchor transaction.
```
LeoPatOZ, line 46:
```suggestion
A corner case arises when an L2 proposer chooses not to include one of the signals stored in the L1 SS within the anchor transaction. This omission can cause any application relying on that signal (as in step 6) to revert. In those cases, the application can always fallback to the 'slow' signalling pathway given the signal can't be deleted and the user can always wait for the next L1 state root. Applications can build mechanisms to incentivize proposers to include their signals in the L2 anchor transaction.
```

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+For the purpose of deriving the storage slot for a signal, we use the [ERC-7201](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/ca7a4e39de0860bbaadf95824207886e6de9fa64/contracts/utils/SlotDerivation.sol#L4) standard formula to create a unique storage slot for each signal that does not collide with other values stored in the same contract. The id of the signal consists of `msg.sender`, `chainId` and `value`.
 
-This approach works well, but has some latency since the latest L1 state root in the case of Taiko is only accessible when the L2 proposer submits a new publication to the L1 and includes a new anchor tx to the L2.
+### L1 -> L2 Same-slot signalling
 
-## Same Slot L1 → L2 Message Passing
+Nethermind's [proposal on same-slot L1 -> L2 message passing](https://ethresear.ch/t/same-slot-l1-l2-message-passing/21186) outlines the concept of signals injected by the L2 proposer into the L2, allowing to bypass the wait for the next L1 state root. The following describes the approach to implement both 'slow' and 'fast' signalling pathways using this minimal rollup stack.
```
ggonzalez94, line 17: I like this section, but I think it should go after `Normal ('slow') Signalling`. The normal signaling should be the first thing that someone reading this doc would learn and then they can be exposed to why we need this other mechanism

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
-3. The inbox contract will verify that the Signals exist in the SS’s storage.
-4. The proposer (owned by the rollup) will call `receiveSignal` on the L2 SS through the anchor transaction executed at the start of L2 blocks. This will fill a mapping of `_receivedSignals` to true (signal —> true)
-5. Any application that requires same slot signalling will just verify the signal is found in the `_receivedSignals` mapping i.e. they don’t need to provide a storage proof
+The simplest form of signalling requires waiting for the next **source chain** state root to be available on the **destination chain** via a trusted mechanism. In the case of L1 → L2 communication, the latest available L1 root is included in the L2 batch by the proposer in an anchor transaction.
 
-An example diagram is provided in the ETH bridge section.
+Using the signal requires the following steps:
+
+1. Store a signal on the L1.
+2. Wait for the next L2 batch to be proposed on the L1.
```
ggonzalez94, line 26: Nit: publication is more consistent with the rest of the repo
```suggestion
2. Wait for the next L2 publication to be proposed on the L1.
```

documentation/SignalServiceDesign.md
```diff
@@ -1,46 +1,81 @@
...
+5. The L2 proposer calls `receiveSignal` on the L2 SS through the anchor transaction executed at the start of L2 blocks. This fills a mapping of `_receivedSignals` to true (signal —> true).
+6. Any application that requires same slot signalling will just verify the signal is found in the `_receivedSignals` mapping (i.e. they don’t need to provide a storage proof).
+
+![Fast Signaling](./fast_signaling.svg)
+
+A corner case is that where an L2 proposer decides not to receive one of the signals stored in the L1 SS within the anchor transaction, making any application using the signal (as in step 6) revert. In those cases, the application can always fallback to the 'slow' signalling pathway given the signal can't be deleted and the user can always wait for the next L1 state root. Applications can build mechanisms to incentivize proposers to include their signals in the L2 anchor transaction.
+
+In the context of an L2 proposer who also issues preconfirmations (either by delegation or any other mechanism), then the inclusion of the signal in the L2 block is guaranteed. Otherwise, the verification at step 4 will fail, reverting the whole L2 batch.
```
ggonzalez94, line 48: I don't think preconfs are needed(they just improve the UX). If you have a deterministic leader election mechanism for L2 sequencers then this will hold true

ggonzalez94 (commented):
Looks good, left a few small comments.
One thing I would like to see documented is [this use case](https://github.com/OpenZeppelin/minimal-rollup/pull/61#issuecomment-2704142525)

documentation/SignalServiceDesign.md
ernestognw: Good question. Inbox is not the best name as it's the one that will go in L2.

Step 3 is done in the L1 and step 4 is done in the L2. However, the misunderstanding relies on **when it happened**. The L2 transaction is built into an L2 block before published on the L1, where the L1 batch includes the L2 transactions (including the anchor).

documentation/SignalServiceDesign.md
ernestognw: True, we should call it always a "publication".

documentation/SignalServiceDesign.md
ernestognw: I updated mentioning the normal passing first. I wouldn't spend too much time on defining the "slow" path since I think it's the "regular" mental model and is described right after this section

ernestognw:
> Looks good, left a few small comments. One thing I would like to see documented is [this use case](https://github.com/OpenZeppelin/minimal-rollup/pull/61#issuecomment-2704142525)

Are you saying that we should outline the specific use case of a "fast-deposit"? My feeling is that it's already implicit by fast tracking signals to the L2.

ggonzalez94:
> Are you saying that we should outline the specific use case of a "fast-deposit"? 

Yes, I do believe this deserves its own section - but I'm fine with merging this PR and then adding it

ggonzalez94 (approved):

ernestognw:
> > Are you saying that we should outline the specific use case of a "fast-deposit"?
> 
> Yes, I do believe this deserves its own section - but I'm fine with merging this PR and then adding it

Disagree, if we were following a structured approach to the L1Rollup, that becomes just a feature on top. The reason this is not clear and we need to write it is because we don't have a proper abstraction yet.

Issue #55, opened 2025-02-28
Implement Delayed Publications


ggonzalez94:
Today we only have an [interface](https://github.com/OpenZeppelin/minimal-rollup/blob/main/src/protocol/taiko_alethia/IDelayedInclusionStore.sol)(should we move it to the main protocol folder?) for delayed publications(a.k.a forced transactions), which is a requirement for Stage1 since the stack will have preconfs.

Some requirements(not exhaustive)
- For now we will support blobs only(no calldata)
- Delayed publications should pay a higher fee than regular publications(since they can be potentially more expensive to prove)
- Proposers are forced to include delayed publications when they publish their next publication(they are posted as a separate publication **after** the proposers' publication)
- Each rollup can have a different deadline for their delayed publications

ggonzalez94:
@LeoPatOZ same here - closing the issue since we already implemented delayed publications

PR #74, opened 2025-03-13
Minimal Delayed Inclusion Store

LeoPatOZ:
Addresses issue https://github.com/OpenZeppelin/minimal-rollup/issues/55. 

- Implements a minimal delayed inclusion store. The basic idea is to have a queue of delayed inclusions that gets cleared once the inbox contract has processed them.
- Inbox contract only has to process the delayed inclusion every X timestamp i.e theres a delay before the next batch of inclusions are processed.


src/protocol/IDelayedInclusionStore.sol
```diff
@@ -6,5 +6,6 @@ interface IDelayedInclusionStore {
         bytes32 blobRefHash;
     }
 
+    /// @notice Returns a list of inclusions that will be processed by the inbox
```
ggonzalez94, line 9:
```suggestion
    /// @notice Returns a list of publications that should be processed by the Inbox
```

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,53 @@
...
+import {IBlobRefRegistry} from "../../blobs/IBlobRefRegistry.sol";
+import {IDelayedInclusionStore} from "../IDelayedInclusionStore.sol";
+
+contract DelayedInclusionStore is IDelayedInclusionStore {
+    // Stores all delayed inclusion requests
+    Inclusion[] private delayedInclusions;
+
+    uint256 private lastInclusionRequestTime;
+
+    uint256 public inclusionDelay;
```
ggonzalez94, line 13: I know this is still a draft, but I would make this variable immutable following what we've been doing with the rest of the codebase and we will at a later point decide which values should be allowed to change and how(storage variable with permissioned access vs function or immutable value that can be upgraded by upgrading the contract)

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,53 @@
...
+        inclusionDelay = _inclusionDelay;
+        inbox = _taikoInbox;
+        blobRefRegistry = IBlobRefRegistry(_blobRefRegistry);
+    }
+
+    /// @dev Registers a blob reference in the blob reference registry
+    /// and stores the hash (Inclusion) in the delayed inclusion queue.
+    /// @param blobIndices An array of blob indices to be registered.
+    function publishDelayed(uint256[] calldata blobIndices) external {
+        (bytes32 refHash,) = blobRefRegistry.registerRef(blobIndices);
```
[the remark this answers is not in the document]
ggonzalez94: In the Inbox we are not getting an already registered blob ref if you check the implementation of `getRef`. It is actually just calculating it in place by calling the `blobhash` function. I think we should do the same here.
(But this also raises the question of who we expect to use the `registerRef` function, since its purpose was to serve as common infra for blob registry, but whoever calls it has to pay an additional amount of gas to store it)

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,53 @@
...
+pragma solidity ^0.8.28;
+
+import {IBlobRefRegistry} from "../../blobs/IBlobRefRegistry.sol";
+import {IDelayedInclusionStore} from "../IDelayedInclusionStore.sol";
+
+contract DelayedInclusionStore is IDelayedInclusionStore {
+    // Stores all delayed inclusion requests
+    Inclusion[] private delayedInclusions;
+
+    uint256 private lastInclusionRequestTime;
```
ggonzalez94, line 11: I don't think a single for the inclusion time works, because it would use the latest time you included something as the effective delay, not the time a publication comes.
This means that if the `inclusionDelay` is 1 hour, and it has been more than that since the time you processed a delayed inclusion if you post a new one it will be included immediately(since the check will pass)

ggonzalez94 (commented):
I believe we need to rethink the timing mechanism for when to pick the delayed inclusions, right now it is based on when the last one was pushed(and not when the publication was added to the queue). An easy way to implement this is to add a `timestamp` to the `Inclusion` struct and then iterate over the delayed publications and return the ones that are due(and then remove them). But gas costs wouldn't be ideal for that solution - I imagine there's a more efficient way to do it

src/protocol/taiko_alethia/DelayedInclusionStore.sol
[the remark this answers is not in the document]
LeoPatOZ: Yeah sorry this is why i was confused as was unsure who calls the `registerRef` function. 
Changed now

src/protocol/taiko_alethia/DelayedInclusionStore.sol
LeoPatOZ: okay changed!

test/DelayedInclusionStore.t.sol
```diff
@@ -0,0 +1,88 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import "forge-std/Test.sol";
+
+import {IDelayedInclusionStore} from "src/protocol/IDelayedInclusionStore.sol";
+import {DelayedInclusionStore} from "src/protocol/taiko_alethia/DelayedInclusionStore.sol";
+import {MockBlobRefRegistry} from "test/mocks/MockBlobRefRegistry.sol";
+
+contract DelayedInclusionStoreTest is Test {
```
ggonzalez94, line 10: I don't think tests are our main focus right now, so I'll make a single comment with some thoughts
1. There's nothing inherently wrong with the `setUp`, but there's something off with the fact that we are going into the future, inserting publications and then coming back, this does not reflect what can actually happen. I believe we should do a process where we insert stuff, warp a few seconds and then insert some other things, but not simulating they were inserted in the future, because that's just not real.
2. We should also check that events are emitted in functions
2. Since we are doing this I would also include tests for `publishDelayed`

ggonzalez94 (commented):
Looks good overall. I left some suggestions

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,96 @@
...
+contract DelayedInclusionStore is IDelayedInclusionStore {
+    struct DueInclusion {
+        bytes32 blobRefHash;
+        uint256 due;
+    }
+
+    /// @notice Emitted when a delayed publication is stored
+    /// @param dueInclusion The delayed publication including
+    /// the blob reference hash and due timestamp
+    event DelayedInclusionStored(DueInclusion dueInclusion);
```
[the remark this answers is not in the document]
LeoPatOZ: done!

test/mocks/MockBlobRefRegistry.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {IBlobRefRegistry} from "../../src/blobs/IBlobRefRegistry.sol";
```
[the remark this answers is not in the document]
LeoPatOZ: Yeah was having issue but all fixed now

test/DelayedInclusionStore.t.sol
LeoPatOZ: Yeah i agree it was more of a quick a dirty test (leftover from measuring gas)

1. I completely revamped the tests should be a lot cleaner now (taken from this article https://sabnock.substack.com/p/take-your-solidity-testing-to-the). hopefully this is a testing framework we can adopt in other tests. 

2. Added tests for the events
3. Added test for publish delayed

test/DelayedInclusionStore.t.sol
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    function test_addOneInclusionOneBlob() public {
+        address sender = address(0x200);
+        uint256[] memory blobIndices = new uint256[](1);
+        // simulate a blob index
+        blobIndices[0] = 1;
+
+        bytes32 expectedRefHash = keccak256((abi.encode(blobReg.getRef(blobIndices))));
+        vm.expectEmit(true, true, false, true);
```
ggonzalez94, line 48: You don't need to explicitely omit the other topics of the event if your event has less than 3 topics
```suggestion
        vm.expectEmit();
```

test/DelayedInclusionStore.t.sol
```diff
@@ -0,0 +1,152 @@
...
+}
+
+contract InclusionAddedStateTest is InclusionAddedState {
+    function test_processDueInclusions_NoneDue() public {
+        vm.prank(inbox);
+        DelayedInclusionStore.Inclusion[] memory inclusions = inclusionStore.processDueInclusions();
+        assertEq(inclusions.length, 0);
+    }
+
+    function test_processDueInclusions_FirstHalfDue() public {
```
ggonzalez94, line 94: Why is this firstHalfDue? I think at this point we only have 25 delayed publications and we are processing all of them

ggonzalez94: Maybe it is just a naming thing from the previous iteration

test/DelayedInclusionStore.t.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    function setUp() public virtual override {
+        super.setUp();
+        vm.warp(waitTime);
+        storeDelayedInclusions(25);
+    }
+}
+
+contract StaggeredInclusionStateTest is StaggeredInclusionState {
+    function test_processDueInclusions_SecondHalfDue() public {
```
ggonzalez94, line 132: I don't think this is correct. When you call `vm.warp` you pass an absolute value, so you are actually moving to timestamp 601(10 minutes + 1 second), you can do `console.log(block.timestamp)` to verify that. So in reality you are processing the first batch of inclusions(the ones you inserted in `InclusionAddedState` `setup`). Since these are "dummy" inclusions they just happen to be the same.
May I recommend merging `StaggeredInclusionStateTest` and `InclusionAddedStateTest` since they don't have too much complexity or functions and just doing the insertions manually in each test(after an initial insertion in the setUp of course if you want). I think that would be clearer

ggonzalez94: this is a bit of a nit: but would also be good to cover a case where you have multiple staggered inclusions to verify the search works well

test/DelayedInclusionStore.t.sol
```diff
@@ -0,0 +1,152 @@
...
+        uint256 baseSlot = uint256(keccak256(abi.encode(slot))) + (index * 2);
+        bytes32 blobRefHash = vm.load(address(inclusionStore), bytes32(baseSlot));
+        uint256 due = uint256(vm.load(address(inclusionStore), bytes32(baseSlot + 1)));
+
+        return DelayedInclusionStore.DueInclusion(blobRefHash, due);
+    }
+}
+
+contract BaseStateTest is BaseState {
+    function test_initialState() public {
```
ggonzalez94, line 37: nit:usually test_SetupState() is [used](https://book.getfoundry.sh/guides/best-practices#general-test-guidance) for the name of this function

test/DelayedInclusionStore.t.sol
```diff
@@ -0,0 +1,152 @@
...
+        assertEq(readInclusionArray(0).blobRefHash, bytes32(0));
+    }
+
+    function test_addOneInclusionOneBlob() public {
```
ggonzalez94, line 41: We've been following the convention: `test_funcName_Description` and on the tests below I see you are also doing that. Would be good to do it on the tests of this contract as well just for consistency

ggonzalez94 (commented):
Almost ready to merge. I left some comments on what inclusions we are testing for, especially on `test_processDueInclusions_SecondHalfDue`.

Also I just noticed that we are merging this to `ahead-of-time-prover` branch, but they are not relayed or have strong dependencies. Should we just merge this to main instead?

test/DelayedInclusionStore.t.sol
ggonzalez94: Yeah, I think this is a nice format. I created an [issue](https://github.com/OpenZeppelin/minimal-rollup/issues/80) to decide how we structure our tests.

test/DelayedInclusionStore.t.sol
ggonzalez94: I think we should also test doing multiple(at least 2) consecutive inclusions and processing them to make sure the head of the queue is updated correctly

test/mocks/MockBlobRefRegistry.sol
```diff
@@ -0,0 +1,39 @@
...
+
+contract MockBlobRefRegistry is IBlobRefRegistry {
```
ggonzalez94, line 6: This implementation is now very similar(the same?) as the actual BlobRefRegistry. Shouldn't we just use the real one? I think the mock makes sense if there's setup and state you don't wan't to deal with on the tests, but not in this case.

ggonzalez94 (commented):
One more comment

test/DelayedInclusionStore.t.sol
LeoPatOZ: updated

test/DelayedInclusionStore.t.sol
LeoPatOZ: updated

test/DelayedInclusionStore.t.sol
LeoPatOZ: Okay i revamped the tests again let me know if this is clearer!

test/mocks/MockBlobRefRegistry.sol
LeoPatOZ: The difference is the:
```solidity
require(blobhashes[i] != 0, "Blob not found");
```
in the real ref registry, i think i would have to mock the actual blob hash which is harder to do but maybe worth it

test/DelayedInclusionStore.t.sol
LeoPatOZ: removed now

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+        DueInclusion memory dueInclusion = DueInclusion(refHash, block.timestamp + inclusionDelay);
+        delayedInclusions.push(dueInclusion);
+        emit DelayedInclusionStored(msg.sender, dueInclusion);
+    }
+
+    /// @inheritdoc IDelayedInclusionStore
+    /// @dev Only returns inclusions if the delay period has passed
+    /// otherwise returns an empty array.
+    /// @dev Can only be called by the inbox contract.
+    function processDueInclusions() external returns (Inclusion[] memory) {
```
ernestognw, line 59:
```suggestion
    function processDueInclusions() public virtual returns (Inclusion[] memory) {
```

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+    /// @param _taikoInbox The address of the Taiko's inbox contract responsible for processing inclusions.
+    constructor(uint256 _inclusionDelay, address _blobRefRegistry, address _taikoInbox) {
+        inclusionDelay = _inclusionDelay;
+        inbox = _taikoInbox;
+        blobRefRegistry = IBlobRefRegistry(_blobRefRegistry);
+    }
+
+    /// @inheritdoc IDelayedInclusionStore
+    /// @dev Stores the blob reference as a DueInclusion
+    function publishDelayed(uint256[] calldata blobIndices) external {
```
ernestognw, line 48: We end up copying blobIndices to memory later so calldata is pointless

```suggestion
    function publishDelayed(uint256[] memory blobIndices) public virtual {
```

ernestognw, line 48:
```suggestion
    function publishDelayed(uint256[] memory blobIndices) public virtual {
```

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+    // Pointer to the first unprocessed element
+    uint256 private head;
+
+    /// @notice The minimum amount of time a delayed publication needs to
+    /// wait in the queue to be included expressed in seconds
+    uint256 public immutable inclusionDelay;
+
+    address public immutable inbox;
+
+    IBlobRefRegistry public immutable blobRefRegistry;
```
ernestognw, line 35: This should be a virtual function imo, perhaps with a default value of a "canonical" blobregistry:

```solidity
function blobRefRegistry() public virtual returns (IBlobRegistry) {
  return IBlobRegistry(<CANONICAL REGISTRY>);
}
```

This is equivalent to an immutable variable since it's setup as part of the code. Also, enables users to override and bring their own contract if they want.

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    /// @notice Emitted when a delayed publication is stored
+    /// @param dueInclusion The delayed publication including
+    /// the blob reference hash and due timestamp
+    /// @param sender The address that stored the delayed publication
+    event DelayedInclusionStored(address indexed sender, DueInclusion dueInclusion);
+
+    /// @notice Emitted when a list of delayed inclusions are processed by the inbox
+    /// @param inclusionsList list of inclusions
+    event DelayedInclusionProcessed(Inclusion[] inclusionsList);
```
ernestognw, line 21: Perhaps these 2 should be in the interface?

Honestly, I'm not sure we even need an interface (this should be an extension to a base inbox contract or similar), but if we do need it, I'd move the events there.

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    // Append-only queue for delayed inclusions
+    DueInclusion[] private delayedInclusions;
```
ernestognw, line 24: Let's denote private variables with a leading underscore

```suggestion
    DueInclusion[] private _delayedInclusions;
```

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    // Pointer to the first unprocessed element
+    uint256 private head;
```
ernestognw, line 27:
```suggestion
    uint256 private _head;
```

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+        emit DelayedInclusionStored(msg.sender, dueInclusion);
+    }
+
+    /// @inheritdoc IDelayedInclusionStore
+    /// @dev Only returns inclusions if the delay period has passed
+    /// otherwise returns an empty array.
+    /// @dev Can only be called by the inbox contract.
+    function processDueInclusions() external returns (Inclusion[] memory) {
+        uint256 len = delayedInclusions.length;
+        if (head >= len) {
```
ernestognw, line 61: We must cache `head` in another variable:

```
uint256 cachedHead = head;
```

Otherwise every further read will be an SLOAD. Probably the compiler optimizations may take care but we still need to update it imo

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+    IBlobRefRegistry public immutable blobRefRegistry;
+
+    /// @param _inclusionDelay The delay before next set of inclusions can be processed.
+    /// @param _blobRefRegistry The address of the blob reference registry contract.
+    /// @param _taikoInbox The address of the Taiko's inbox contract responsible for processing inclusions.
+    constructor(uint256 _inclusionDelay, address _blobRefRegistry, address _taikoInbox) {
+        inclusionDelay = _inclusionDelay;
+        inbox = _taikoInbox;
+        blobRefRegistry = IBlobRefRegistry(_blobRefRegistry);
+    }
```
ernestognw, line 44: I'd remove the constructor if all these variables become public virtual functions

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,100 @@
...
+            } else {
+                r = mid;
+            }
+        }
+
+        // l now is the first index where the inclusion is not yet due.
+        uint256 count = l - head;
+        if (count == 0) {
+            return new Inclusion[](0);
+        }
```
ernestognw, line 82: I would abstract this into an internal function (e.g. `_countDelayedInclusions`). In my opinion, if anyone wants to rely on a mapping instead they should be able to do it by overriding `_countDelayedInclusions`

Perhaps:

```solidity
function _countDelayedInclusions() internal virtual returns (uint256 last, uint256 count) {
        uint256 l = head;
        uint256 r = len;
        uint256 timestamp = block.timestamp;
        while (l < r) {
            uint256 mid = (l + r) >> 1;
            if (delayedInclusions[mid].due <= timestamp) {
                // If inclusion at mid is due, search to the right.
                l = mid + 1;
            } else {
                r = mid;
            }
        }

        return (l, l - head);
}
```

test/mocks/MockBlobRefRegistry.sol
LeoPatOZ: removed the mock blob ref registry

LeoPatOZ:
Changed base to main

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,103 @@
...
+    struct DueInclusion {
+        bytes32 blobRefHash;
+        uint256 due;
+    }
+
+    /// @notice Emitted when a delayed publication is stored
+    /// @param dueInclusion The delayed publication including
+    /// the blob reference hash and due timestamp
+    /// @param sender The address that stored the delayed publication
+    event DelayedInclusionStored(address indexed sender, DueInclusion dueInclusion);
```
ggonzalez94, line 23: Just curious why this event is on the contract and not the interface like `DelayedInclusionProcessed`? Seems natural that if we have both functions in the interface the two events should also be together?

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -0,0 +1,103 @@
...
+        emit DelayedInclusionStored(msg.sender, dueInclusion);
+    }
+
+    /// @inheritdoc IDelayedInclusionStore
+    /// @dev Only returns inclusions if the delay period has passed
+    /// otherwise returns an empty array.
+    /// @dev Can only be called by the inbox contract.
+    function processDueInclusions() external returns (Inclusion[] memory) {
+        uint256 len = _delayedInclusions.length;
+        uint256 cachedHead = _head;
```
ggonzalez94, line 63: very minor nit: we don't tend to use `cachedVariableName` and I haven't seen this as a common practice in solidity. Maybe we can just name it head?

ggonzalez94 (approved):
LGTM! Just left a couple final small suggestions, but it is ready to be merged IMO

src/protocol/taiko_alethia/DelayedInclusionStore.sol
LeoPatOZ: changed

test/DelayedInclusionStore.t.sol
```diff
@@ -0,0 +1,213 @@
...
+
+        vm.blobhashes(blobHashes);
+        bytes32 expectedRefHash = keccak256((abi.encode(blobReg.getRef(blobIndices))));
+        vm.expectEmit();
+        emit DelayedInclusionStore.DelayedInclusionStored(
+            sender, DelayedInclusionStore.DueInclusion(expectedRefHash, vm.getBlockTimestamp() + inclusionDelay)
+        );
+        vm.prank(sender);
+
+        vm.blobhashes(blobHashes);
```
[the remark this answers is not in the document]
LeoPatOZ: oops thats a mistake

src/protocol/taiko_alethia/DelayedInclusionStore.sol
ggonzalez94: Makes sense!

nikeshnazareth:
Looks good. I'm approving the PR, but I'll leave these thoughts for your consideration:
- I think it's weird not to check `msg.sender == inbox` right at the start. I think reverting is correct behaviour if it's called by the wrong contract, even if there are no state changes. Also, it's generally not good practice to leak information to unauthorised users (eg. if you try to open a letterbox with the wrong key, it should fail to open whether or not there are letters inside). That doesn't exactly translate to this scenario since it's all public information anyway, but failing to enforce access control in some cases does seem more fragile anyway.
- Do you think we should use a ring buffer to overwrite used records?
- The binary search seems excessive to me, because I'd guess there are only a handful of delayed publications at any point in time. But maybe I'm wrong about that, and it doesn't hurt, even if it's slightly overengineered.
- I think I agree with Ernesto's point that we should think of this as a delayed inclusion behaviour of the inbox (so it's an abstract ancestor contract rather than a whole separate contract) but that's probably part of a larger restructure anyway.

ggonzalez94:
> I think it's weird not to check msg.sender == inbox right at the start.

I proposed this optimization since like you said the information is already public and in the case of empty list of due incluse(which should happen often) you save some gas. But agree it might be a bit of a micro optimization.

> The binary search seems excessive to me, because I'd guess there are only a handful of delayed publications at any point in time. But maybe I'm wrong about that, and it doesn't hurt, even if it's slightly overengineered.

We had some discussions around this in slack, and I was a bit hesitant of the binary search(even if I initially recommended it lol) knowing that the queue will almost always be small. But since the cost of cold storage access dwarfs the rest, binary search is likely the most gas-efficient approach for almost any queue size above 2-3 elements. And for a smaller size, the additional computation cost should still be fairly minimal.

> Do you think we should use a ring buffer to overwrite used records?

I think this makes sense definitely. We should just define what is the max size of the queue and prevent new delayed publications if the queue is full?

Issue #67, opened 2025-03-07
Imperfect atomic execution


nikeshnazareth:
## Overview
- I had some thoughts (two incremental ones and one interesting one) that could improve the flexibility of rollups
- They seem to work well together
- I haven't really thought them through in detail. I just want to put the ideas on the table so we can riff on them.

## Include post-state in publication

The first incremental idea is that I think the proposer (excluding the delayed inclusion publications) should include a claimed L2 state root (ie. the checkpoint after applying this publication) within the publication as one of the attributes. The whole publication should be considered invalid if it does not result in this root.

The idea is just to exploit an existing asymmetry:

- there is a very valuable piece of information (the new L2 state root) that users want to know
- the proposer already knows it and it would be trivial for them to tell us
- we should give them a strong incentive to do so

This is, of course, not a proof that the publication is valid but some observers might consider it enough. I think this is just an extension of the insight that [security is a property of the confirmation rule](https://dba.xyz/do-rollups-inherit-security/#confirmation-rules). Depending on your use case, you might require a transaction to be finalized, or 6 blocks deep, or optimistically "proven", or just proposed, or preconfirmed,  or in the mempool, or whatever. In all cases you're comparing the probability that this is a reliable indicator against the value of whatever you're doing. This suggestion is a simple and cheap way of creating a pretty strong signal.

## Make invalid post-states a slashable offence

This is an extension of the same idea. We have already assumed a bunch of infrastructure is in place to encourage proposers to make preconfirmations. We can reuse it to encourage everything else we want them to do. I also think that binding their stake to group properties like the state of the whole chain can reduce the scope for shenanigans.

In this case, it could be something like:

- if the claimed root is invalid, the proposer's stake is slashed
- in addition to providing their own state root, the next proposer has to indicate whether they're building on the previous one (or whether they consider it to be a no-op). In this way, agreement/disputes are automatically recorded in order to increase the strength of the signal. Perhaps the next proposer should receive some of the slashed stake on a successful dispute.
   - I realise this sounds like the optimistic rollup mechanism, but it doesn't change the fact that actual proofs are still required.
   - I wouldn't suggest this if the proposer weren't already staked (because of centralisation concerns) but since we've already accepted that condition, we might as well include the optimistic signal while waiting for the proof.

We should consider if there are other properties that could also be slashable offences. I have a tentative design for a partial atomic execution mechanism, but it requires a new transaction type and it might not be necessary if the next idea makes sense.

[...]

PR #78, opened 2025-03-20
preemptive provable assertions

nikeshnazareth:
Draft PR to explain the idea in #67


xiaodino (commented):
The current implementation focuses on making the block invalid if assertions fail, which is great for security. One thing to consider is whether the proposer's stake gets slashed in that event (since this is similar to an invalid block proposal)​.

This might be handled outside of this PR (perhaps by the Proposer/Prover market contract), but it’s part of the bigger picture: preemptive assertions provide a signal that a block is likely valid; if that signal is false, there should be consequences. Right?

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -61,10 +64,21 @@ contract TaikoInbox is IInbox {
         });
         require(metadata.anchorBlockHash != 0, "blockhash not found");
 
-        bytes[] memory attributes = new bytes[](3);
+        // Perform preemptively asserted queries
+        uint256 nQueries = queries.length;
+        uint256[] memory results = new uint256[](nQueries);
+        for (uint256 i = 0; i < nQueries; i++) {
+            require(queries[i].destination != address(publicationFeed), "Cannot query publication feed");
+            (bool success, bytes memory returndata) = queries[i].destination.call(queries[i].callData);
```
xiaodino, line 72: For safety, should we use ```.staticcall``` instead of ```.call``` when executing L1 queries in ```TaikoInbox.publish``` here? We only need to read data.

src/protocol/taiko_alethia/preemptive_assertions/PreemptiveProvableAssertions.sol
```diff
@@ -0,0 +1,63 @@
...
+
+contract PreemptiveProvableAssertions is
+    L1QueriesPublicationTime,
+    RealtimeL1StorageRead,
+    L2QueryFutureBlock,
+    CrossRollupStorageRead
+{
+    constructor(address taikoAnchor) PreemptiveProvableAssertionsBase(taikoAnchor) {}
+
+    function resolveAssertions(bytes32 attributesHash, bytes calldata proofs) external onlyAnchor {
```
xiaodino, line 19: Quick question on the gas usage of `resolveAssertions(bytes32 attributesHash, bytes calldata proofs)`:

It seems the gas cost could grow with the number and complexity of assertions. How do we plan to balance this practically? Is there a recommended maximum number of assertions per publication, or perhaps guidelines for keeping assertion logic lightweight?

PR #85, opened 2025-03-25
Use State structure for tests

### LINKED ISSUE #80: Decide a test structure
So far we've been fairly loose in how we organize tests, mostly putting them under a single file per contract. We should make a conscious decision that improves organization, but is not overkill(if in doubt we should always defer to the simplest structure).

## Test organization
@LeoPatOZ suggested using this [state driven](https://sabnock.substack.com/p/take-your-solidity-testing-to-the) approach. Foundry has also some [guidance](https://book.getfoundry.sh/guides/best-practices#tests) on how to best organize your tests.

nikeshnazareth:
Closes #80.

I wanted to try Leo's suggestion for a state-driven approach. It seems to be much more flexible and easier to reason about. I'm in the middle of migrating the existing tests to this structure, but I ran out of time. I will continue updating it this week but I thought it would be useful to publish the current version in case you have any feedback. Note:
- this is based off PR 49 but I do not think this PR should be a blocker for merging that one.
- I know we have not decided to use this structure. I'm just experimenting with it.

EDIT: this is obsolete and should be replaced by #134


nikeshnazareth:
Closing because it is superseded by #134

PR #88, opened 2025-03-27
Add a basic style guide and contribution guidelines

ggonzalez94:
- A minimal set of conventions help keep our contracts consistent, and ensure our PR discussions are focused on the code changes, and not the style. It also reduces some of the mental overhead of taking basic code decisions.
- We didn't have a contribution guideline. I don't think we are a stage that we want to have many external contributors, but still good to have this sooner rather than later.


STYLE_GUIDE.md
```diff
@@ -0,0 +1,132 @@
...
+
+1. Event names should be in the past tense.
+
+    Events should track things that happened and so should be past tense. Using past tense also helps avoid naming collisions with structs or functions.
+
+    ```solidity
+    event Published();
+    ```
+
+1. Constructor parameters, when they overshade a state variable, should be prefixed with an underscore.
```
LeoPatOZ, line 32: I think this might contradict rule 1. If you have a state variable called `_totalSupply` then:
```solidity
    uint256 private _totalSupply;
    constructor(uint256 _totalSupply) {
        totalSupply = _totalSupply;
    }
```
will fail as its 'shadowing an existing variable'

LeoPatOZ (approved):
LGTM 💯

ggonzalez94:
> Also i dont to mention here or not but any preference for require statement vs modifiers?

I usually prefer require/if statements since it is easier to read(vs having to go to the modifier). But when the check is repeated multiple times it does make sense to abstract in a modifier

PR #89, opened 2025-03-27
Signalling

LeoPatOZ:
This PR aims at adding additional signalling functionality.

1. Adding a canonical ETH bridge
2. Simplifying the signal service
3. Added historical commitment tracking


src/libs/LibSignal.sol
```diff
@@ -0,0 +1,85 @@
...
+/// @dev Library for secure broadcasting (i.e. signaling) cross-chain arbitrary data.
+///
+/// Signaling a value consists of storing a `bytes32` in a namespaced storage location to guarantee non-collision
+/// slots derived by EVM languages such as Solidity or Vyper. Smart contracts utilizing this library will signal values
+/// with the `signal` function, allowing to generate a storage proof with an `eth_getProof` RPC call.
+///
+/// Later, on a destination chain the signal can be proven by providing the proof to `verifySignal` as long as the
+/// state root is trusted (e.g. the L1 state root can made available on the L2 by the proposer).
+library LibSignal {
+    using SafeCast for uint256;
```
ggonzalez94, line 18: Minor optimization: I don't think the risk of overflowing when casting the chainId is significant, so maybe we can do the conversion directly instead of using the safe version that does an extra check

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+    /// @param value Data to be stored (signalled)
+    function sendSignal(bytes32 value) external returns (bytes32 slot);
+
+    /// @dev Checks if a signal has been stored
+    /// @dev Note: This does not been it has been 'sent' to destination chain.
+    /// @param value Value to be checked is stored
+    function isSignalStored(bytes32 value) external view returns (bool);
+
+    /// @dev Verifies if the signal can be proved to be part of a merkle tree
+    /// defined by `root` for the specified signal service storage.
```
ggonzalez94, line 38: Signals can be verified multiple times correct? Perhaps the natspec should make this clear? e.g.
`/// @dev Signals are not deleted when verified, and can be verified multiple times by calling this function`

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+    error SignalNotReceived(bytes32 value, bytes32 root);
+
+    /// @dev Stores a data signal and returns its storage location.
+    /// @param value Data to be stored (signalled)
+    function sendSignal(bytes32 value) external returns (bytes32 slot);
+
+    /// @dev Checks if a signal has been stored
+    /// @dev Note: This does not been it has been 'sent' to destination chain.
+    /// @param value Value to be checked is stored
+    function isSignalStored(bytes32 value) external view returns (bool);
```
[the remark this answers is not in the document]
LeoPatOZ: yeah i agree!

src/libs/LibSignal.sol
LeoPatOZ: i agree removed now

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+    /// @param value Value to be verified
+    /// @param accountProof Merkle proof for account state against global stateRoot
+    /// @param stateProof Merkle proof for slot value against account's storageRoot
+    function verifySignal(
+        bytes32 root,
+        uint64 chainId,
+        bytes32 value,
+        bytes[] memory accountProof,
+        bytes[] memory stateProof
+    ) external;
```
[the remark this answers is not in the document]
LeoPatOZ: 1. Yes i removed it, I'm just unsure how to get this trusted root atm
2. Fixed the comment
3. correct fixed that now

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+///
+/// Signals enable generalized on-chain communication, primarily for data transmission rather than value transfer.
+/// Applications can leverage signals to transfer value through secondary implementations.
+///
+/// Signals are broadcast without specific recipients, allowing flexible cross-chain data sourcing from any
+/// source chain state (e.g., full transaction logs or contract storage).
+interface ISignalService {
+    /// @dev Emitted when a signal is sent.
+    /// @param signal Signal value that was stored
+    event SignalSent(bytes32 signal);
```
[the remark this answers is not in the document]
LeoPatOZ: yeah updated the event should be clearer now

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+pragma solidity ^0.8.28;
+
+/// @title ISignalService
+/// @dev Secure cross-chain messaging system for broadcasting arbitrary data (i.e. signals).
+///
+/// Signals enable generalized on-chain communication, primarily for data transmission rather than value transfer.
+/// Applications can leverage signals to transfer value through secondary implementations.
+///
+/// Signals are broadcast without specific recipients, allowing flexible cross-chain data sourcing from any
+/// source chain state (e.g., full transaction logs or contract storage).
```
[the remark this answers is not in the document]
LeoPatOZ: fixed

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+/// Signals enable generalized on-chain communication, primarily for data transmission rather than value transfer.
+/// Applications can leverage signals to transfer value through secondary implementations.
+///
+/// Signals are broadcast without specific recipients, allowing flexible cross-chain messaging
+interface ISignalService {
+    /// @dev Emitted when a signal is sent.
+    /// @param chainId The chain ID of the source chain where the signal was sent
+    /// @param sender The address that sent the signal on the source chain
+    /// @param value The signal value
+    event SignalSent(address sender, uint256 chainId, bytes32 value);
```
ggonzalez94, line 16: For easier off-chain search I think sender should be indexed. Also does it make sense to include chainId as part of the event? Anyone listening to this event already knows the chainId
```suggestion
    event SignalSent(address indexed sender, , bytes32 value);
```

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    /// @dev Emitted when a signal is verified.
+    /// @param chainId The chain ID of the source chain where the signal was sent
+    /// @param value Value that was signaled
+    event SignalVerified(uint64 chainId, address sender, bytes32 value);
```
ggonzalez94, line 21:
```suggestion
    event SignalVerified(uint64 indexed chainId, address indexed sender, bytes32 value);
```

src/protocol/ICheckpointSyncer.sol
```diff
@@ -0,0 +1,37 @@
...
+
+/// @dev Synchronize checkpoints from different chains
+///
+/// A checkpoint is any value (typically a state root) that uniquely identifies the state of a chain at a
+/// specific height (i.e. an incremental identifier like a blockNumber, publicationId or even a timestamp).
+/// Only a trusted syncer can sync checkpoints. For example, only the `CheckpointTracker` can sync roots on the L1 and
+/// the anchor can sync blockhash on the L2.
+interface ICheckpointSyncer {
+    /// @dev A new `checkpoint` has been synced for `chainId` at a specified `height`.
+    event CheckpointSynced(bytes32 checkpoint, uint64 chainId, uint256 height);
```
ggonzalez94, line 14: Same maybe chainId and height should be indexed?

ggonzalez94 (commented):
I know this is still in draft, but I like how it is looking, much cleaner!
One question: Do you think it is worth implementing the ring buffer mechanism now or do you prefer to implement that later?

src/protocol/CheckpointSyncer.sol
```diff
@@ -0,0 +1,47 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {ICheckpointSyncer} from "./ICheckpointSyncer.sol";
+import {ICheckpointTracker} from "./ICheckpointTracker.sol";
+
+abstract contract CheckpointSyncer is ICheckpointSyncer {
+    address private _trustedSyncer;
+
+    mapping(uint256 height => bytes32 checkpoint) private _checkpoints;
```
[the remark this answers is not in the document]
LeoPatOZ: Yeah i was in between the two, its just in my mind a commitment is an unproven checkpoint and this only tracks "proven checkpoints" aka block hash / state root but yes confusing with our custom checkpoint struct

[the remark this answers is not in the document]
LeoPatOZ: changed everything to reflect commitment, also changed contract to commitment store as syncing is probably the wrong verb

src/protocol/CheckpointSyncer.sol
```diff
@@ -0,0 +1,47 @@
...
+    function syncCheckpoint(uint256 height, bytes32 checkpoint) external virtual onlyTrustedSyncer {
+        require(checkpoint != bytes32(0), "Checkpoint cannot be 0");
+        //TODO: Should we check height > previous height? would require more storage
+        _checkpoints[height] = checkpoint;
+        emit CheckpointSynced(checkpoint, uint64(block.chainid), height);
+    }
+
+    /// @dev Internal helper to validate the trusted syncer.
+    function _checkTrustedSyncer(address caller) internal view {
+        require(caller == _trustedSyncer, "UnauthorizedCheckpointTracker");
```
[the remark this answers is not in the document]
LeoPatOZ: changed

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,51 @@
...
+    /// @param value Value to be checked is stored
+    /// @param sender The address that sent the signal
+    function isSignalStored(bytes32 value, address sender) external view returns (bool);
+
+    /// @dev Verifies if the signal can be proved to be part of a merkle tree
+    /// @dev Signals are not deleted when verified, and can be
+    /// verified multiple times by calling this function
+    /// @dev see `LibSignal.verifySignal`
+    /// @dev Height refers to the block number / commitmentId where the trusted root is mapped to
+    function verifySignal(
```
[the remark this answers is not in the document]
LeoPatOZ: true changed it now

LeoPatOZ:
> I know this is still in draft, but I like how it is looking, much cleaner! One question: Do you think it is worth implementing the ring buffer mechanism now or do you prefer to implement that later?

added the ring buffer now

src/libs/LibSignal.sol
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /// @dev Signal a `value` at a namespaced slot for the current `msg.sender` and `block.chainid`.
+    function signal(bytes32 value) internal returns (bytes32) {
+        return signal(msg.sender, value);
+    }
+
+    /// @dev Signal a `value` at a namespaced slot for the current `block.chainid`.
+    function signal(address account, bytes32 value) internal returns (bytes32) {
+        return signal(uint64(block.chainid), account, value);
```
KumaCrypto, line 45: Hey! I'd prefer if we used safe typecasting for `chainid`. Without it, if `chainid` exceeds `type(uint64).max`, it could get truncated and mistakenly represent a different chain. There's [EIP-2294](https://eips.ethereum.org/EIPS/eip-2294), but it hasn't been accepted in a long time, and some chains might use large values for `chainid`. Wdyt?

src/libs/LibSignal.sol
```diff
@@ -0,0 +1,85 @@
...
+
+    /// @dev Signal a `value` at a namespaced slot for the current `block.chainid`.
+    function signal(address account, bytes32 value) internal returns (bytes32) {
+        return signal(block.chainid.toUint64(), account, value);
+    }
+
+    /// @dev Signal a `value` at a namespaced slot. See `deriveSlot`.
+    function signal(uint64 chainId, address account, bytes32 value) internal returns (bytes32) {
+        bytes32 slot = deriveSlot(chainId, account, value);
+        slot.getBytes32Slot().value = value;
```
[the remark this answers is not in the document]
KumaCrypto: Hey, the implementation is:
```solidity
    function getBytes32Slot(bytes32 slot) internal pure returns (Bytes32Slot storage r) {
        assembly ("memory-safe") {
            r.slot := slot
        }
    }
```

src/libs/LibSignal.sol
LeoPatOZ: We originally had safe cast for chain id but i think we decided its not needed (i think we're always dealing with small chain ids)  @ggonzalez94

KumaCrypto: Interesting, but I think it is better to be on the safe side : )

ggonzalez94: Thanks for this, do you know any chain that realistically has chainIds bigger than `type(uint64).max`? I'm not too familiar with the process of assigning chainIds.

src/protocol/CommitmentStore.sol
```diff
@@ -0,0 +1,83 @@
...
+        uint256 _bufferLength = bufferLength();
+        require(latestHeight_ - height <= _bufferLength, CommitmentNotFound());
+        uint256 index = height % _bufferLength;
+        return _commitments[index];
+    }
+
+    /// @inheritdoc ICommitmentStore
+    function storeCommitment(uint256 height, bytes32 commitment) external virtual onlyAuthorizedCommitter {
+        // CHECK: These could be skipped as caller always checks them
+        require(commitment != bytes32(0), "Commitment cannot be 0");
```
ggonzalez94, line 69: Yes, I agree the store should be a "dumb store" that just saves whatever commitment the caller passes without much validation

src/protocol/taiko_alethia/SignalService.sol
ggonzalez94, line 1: I believe the SignalService should be part of the protocol, correct? I don't think it has anything specific or opinionated about Taiko in particular. Also if the ETHBridge uses it(and the ETHBridge is part of the protocol), it makes sense that the SS is also part of the protocol

src/protocol/taiko_alethia/SignalService.sol
```diff
@@ -0,0 +1,81 @@
...
+import {LibTrieProof} from "../../libs/LibTrieProof.sol";
+import {CommitmentStore} from "../CommitmentStore.sol";
+import {ETHBridge} from "../ETHBridge.sol";
+import {ISignalService} from "../ISignalService.sol";
+
+/// @dev Implementation of a secure cross-chain messaging system for broadcasting arbitrary data (i.e. signals).
+///
+/// The service defines the minimal logic to broadcast signals through `sendSignal` and verify them with
+/// `verifySignal`. The service is designed to be used in conjunction with the `ETHBridge` contract to
+/// enable cross-chain communication and the CommitmentStore to retrieve trusted state roots.
```
ggonzalez94, line 14: Nit: I think all the other interfaces and abstract contracts define what each component should do, so I would focus the description of this contract to explain what the whole thing put together does(send + verify + bridge native token)
```suggestion
/// @dev Implementation of a secure cross-chain messaging system for broadcasting and verifying arbitrary data (i.e. signals), and bridging the native token(ETH)
```

src/protocol/taiko_alethia/SignalService.sol
```diff
@@ -0,0 +1,81 @@
...
+
+    function _verifySignal(
+        uint64 chainId,
+        uint256 height,
+        address sender,
+        bytes32 value,
+        bytes[] memory accountProof,
+        bytes[] memory storageProof
+    ) internal view {
+        // WARN: commitmentAt(height) might not be the 'state root' of the chain
```
ggonzalez94, line 75: I believe this is more a TODO than a warning. If we wanted to make this truly generic we would accept a single `bytes calldata proof` and then decode it into a known struct + make the function virtual.
I don't think we need to go trough all that trouble right now, we just need to keep in mind for later:
1. On the L1 we are storing the state root of the L2, so I think we are good
2. On the L2 we are storing the block hash of the L1, so we need to do something. Probably the best thing to do is that the anchor stores the state root in the `CommitmentStore` instead
3. Do we want to make it easy for people to use something different than the state root as a commitment? e.g. the storage root of a single contract

src/protocol/ETHBridge.sol
```diff
@@ -0,0 +1,76 @@
...
+import {IETHBridge} from "./IETHBridge.sol";
+import {StorageSlot} from "@openzeppelin/contracts/utils/StorageSlot.sol";
+
+/// @dev Abstract ETH bridging contract to send native ETH to other chains using storage proofs.
+///
+/// IMPORTANT: No recovery mechanism is implemented in case an account creates a deposit that can't be claimed.
+abstract contract ETHBridge is IETHBridge {
+    using StorageSlot for bytes32;
+
+    mapping(bytes32 id => bool) _claimed;
```
ggonzalez94, line 15: By default it would be internal. I imagine we are actually ok with `private`, but in any case we should make it explicit
```suggestion
    mapping(bytes32 id => bool claimed) private _claimed;
```

src/protocol/ETHBridge.sol
```diff
@@ -0,0 +1,76 @@
...
+        emit ETHDepositClaimed(id, deposit);
+    }
+
+    /// @dev Function to transfer ETH to the receiver but ignoring the returndata.
+    /// @param to Address to send the ETH to
+    /// @param value Amount of ETH to send
+    /// @param data Data to send to the receiver
+    function _sendETH(address to, uint256 value, bytes memory data) internal virtual returns (bool success) {
+        assembly ("memory-safe") {
+            // CHECK: use staticcall?
```
ggonzalez94, line 65: Why would you want to do it? Isn't it for read only calls?

src/protocol/ETHBridge.sol
```diff
@@ -0,0 +1,76 @@
...
+            success := call(gas(), to, value, add(data, 0x20), mload(data), 0, 0)
```
ggonzalez94, line 66: I know we've used assembly to send eth on other contracts, e.g. the ProverManager, but I am a bit hesitant of how much of a saving it really is vs the mental complexity it adds. Also I've seen most compilers should be able to produce almost identical bytecode. Do you think it is worth changing it to:
`(bool success, ) = to.call{value: value, gas: gasleft()}(data);`

src/protocol/IETHBridge.sol
```diff
@@ -0,0 +1,67 @@
...
+
+    /// @dev ETH Deposit identifier.
+    /// @param deposit The ETH deposit
+    function getDepositId(ETHDeposit memory deposit) external view returns (bytes32 id);
+
+    /// @dev Creates an ETH deposit with `msg.value` for the receiver (`to`) to be claimed on the `chainId`.
+    /// @param chainId The destination chain id
+    /// @param to The receiver of the deposit
+    /// @param data Any calldata to be sent to the receiver in case of a contract
+    function depositETH(uint64 chainId, address to, bytes memory data) external payable returns (bytes32 id);
```
ggonzalez94, line 53: nit: all the other functions don't have ETH in their name. maybe this should also just be deposit
```suggestion
    function deposit(uint64 chainId, address to, bytes memory data) external payable returns (bytes32 id);
```

ggonzalez94 (commented):
Just did another pass, left a few suggestions. The two major things I believe are:
1. Fixing the ring buffer in the `CommitmentStore`
2. We should add at least some basic tests before considering merging this

src/protocol/ETHBridge.sol
LeoPatOZ: yes true i think i was just following the previous impl

src/protocol/CommitmentStore.sol
```diff
@@ -0,0 +1,83 @@
...
+import {ICommitmentStore} from "./ICommitmentStore.sol";
+
+/// @dev Base contract for storing commitments.
+/// Uses a circular buffer to store commitments, size is defined by `bufferLength`.
+abstract contract CommitmentStore is ICommitmentStore {
+    address private _authorizedCommitter;
+    address private immutable _rollupOperator;
+    uint256 private _latestHeight;
+
+    mapping(uint256 index => bytes32 commitment) private _commitments;
```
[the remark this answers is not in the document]
LeoPatOZ: I definitely think we should have a ring buffer but i agree maybe we just keep track of it in a separate PR. I forgot that heights aren't sequential :(

src/protocol/CommitmentStore.sol
```diff
@@ -0,0 +1,83 @@
...
+pragma solidity ^0.8.28;
+
+import {ICheckpointTracker} from "./ICheckpointTracker.sol";
+import {ICommitmentStore} from "./ICommitmentStore.sol";
+
+/// @dev Base contract for storing commitments.
+/// Uses a circular buffer to store commitments, size is defined by `bufferLength`.
+abstract contract CommitmentStore is ICommitmentStore {
+    address private _authorizedCommitter;
+    address private immutable _rollupOperator;
```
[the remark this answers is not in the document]
LeoPatOZ: The issue is this circular dependency if we add the authorizedCommitter in the constructor.  As now the checkpoint tracker / anchor needs to the SS in the constructor and the SS needs the  checkpoint tracker / anchor in its constructor. 
Ive changed it to be ownable for now (option 1)

src/protocol/ISignalService.sol
```diff
@@ -0,0 +1,55 @@
...
+    event SignalSent(address indexed sender, bytes32 value);
+
+    /// @dev Emitted when a signal is verified.
+    /// @param chainId The chain ID of the source chain where the signal was sent
+    /// @param value Value that was signaled
+    event SignalVerified(uint64 indexed chainId, address indexed sender, bytes32 value);
+
+    /// @dev Error when a signal fails to be verified.
+    /// @param value Value that was not verified
+    error SignalNotReceived(bytes32 value);
```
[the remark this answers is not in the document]
LeoPatOZ: yes probably!

src/protocol/CommitmentStore.sol
```diff
@@ -0,0 +1,83 @@
...
+    /// @inheritdoc ICommitmentStore
+    function latestHeight() public view virtual returns (uint256) {
+        return _latestHeight;
+    }
+
+    /// @inheritdoc ICommitmentStore
+    function commitmentAt(uint256 height) public view virtual returns (bytes32) {
+        uint256 latestHeight_ = _latestHeight;
+        require(height <= latestHeight_, HeightGreaterThanLatest());
+        uint256 _bufferLength = bufferLength();
```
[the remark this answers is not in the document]
LeoPatOZ: removed circular buffer for now

src/protocol/CommitmentStore.sol
LeoPatOZ: removed

src/protocol/taiko_alethia/SignalService.sol
LeoPatOZ: yeah makes sense!

src/libs/LibSignal.sol
KumaCrypto: I don’t know of any chainids higher than `type(uint64).max`. 

That said, there isn’t a strict restriction on chainids at the moment -- aside from the indirect limit imposed by EIP-155. This EIP defines the `v` value in a signature as `{0,1} + CHAIN_ID * 2 + 35`, which means the maximum allowable chainid  is `(2^256 - 37) / 2`. If the chainid exceeds that, EIP-155 won’t be usable on that blockchain.

So my point is: if there’s no enforced limit, it’s technically possible : )

src/protocol/IETHBridge.sol
LeoPatOZ: Yeah im not sure what the best convention is. OP seems to use ETH in the function [name](https://github.com/ethereum-optimism/optimism/blob/62c7f3b05a70027b30054d4c8974f44000606fb7/packages/contracts-bedrock/contracts/universal/StandardBridge.sol#L209) but other chains leave it generic like 'sendMessage'. I think i prefer to be more explicit especially when there might be other operations like depositERC20. Maybe i can change the claim to claimETHDeposit?

src/libs/LibSignal.sol
LeoPatOZ: Okay changed it to safecast, i dont think theres any overhead adding it tbh

src/protocol/taiko_alethia/SignalService.sol
LeoPatOZ: okay changed is this what you had in mind ?

src/protocol/taiko_alethia/SignalService.sol
LeoPatOZ: I addressed point 3. The old taiko implementation assumes that the `root` is a contract root if the account proof is empty, so now its slightly more generic as to what root means. Also made the proof a memory bytes as mentioned which gets decoded into a struct. 
We still need to address point 1 and 2

src/protocol/IETHBridge.sol
ggonzalez94: I think the contract you are interacting with already gives the context, so I prefer to keep it simple. But I don't have a strong opinion, so you can pick whichever makes the most sense to you - we should just keep it consistent across functions in the same contract

nikeshnazareth:
Thanks for creating this! The storage tests/proofs seem pretty powerful. Nice.

There are a few things that don't make sense to me or look like bugs. 

I think it might make sense to walk through my understanding (based on my reading of the code) in small steps to make sure we're on the same page. Can you confirm if all of this is correct:
- firstly, the `signal-service.md` documentation is no longer accurate
- we are expecting to deploy a global (singleton) `SignalService` contract on both L1 and L2
- it contains two distinct storage patterns (I'm ignoring `_claimed` because that's obvious):
   - the `_commitments` mapping is a privileged place that's used to keep validated state roots of the other chain (ignoring the state root / block hash distinction for now). So on L1 it will have L2 checkpoints and on L2 it will have L1 anchor blocks.
   - signals can be posted by anyone to a pseudorandom location based on the sender/value
- when sending a signal on the source chain, the value is saved in the `SignalService` contract on the source chain
- when verifying a signal on the destination chain, the value is checked against the commitment saved in the destination chain. This means we're proving the signal existed in the state root on the other chain. However, the relevant `account` is `address(this)` (ie. the address of the `SignalService` on the destination chain). For this to work, we assume that the signal service is deployed at the same address on both chains. Is that right? How do we intend to make that happen? Are we hardcoding this value into the rollups?
- the height of a signal doesn't matter. We use it to identify which state root to check, but any state root that has the signal is fine. 
- An ETH deposit is achieved by sending ETH to the `SignalService`, where it is locked, and a corresponding signal is saved. On the other chain, the signal is verified, and the funds are sent out from that contract.
- As I understand it, this implies we've preloaded the `SignalService` on the L2 with a bunch of ETH so that it already has funds to facilitate the claim on L2.
- An ETH withdrawal can go through the same logic. It's just a deposit on L2 that is claimed on L1. This works because the only funds (outside the `SignalService`) on L2 should have come from a previous bridge transfer, so all the L2 funds are escrowed in the L1 `SignalService`

At a high level this makes sense to me, but there are some things that seem weird:
- When the `CheckpointTracker` saves a proven checkpoint, and then immediately stores it in the `SignalService`, this seems entirely redundant. We should save it in just one place.
- My understanding of why we're syncing the checkpoints to a separate `SignalService` is because we have multiple trackers (at least one per rollup) and we want a central place where everything is saved. But in that case:
   - we should use role-based access control so multiple trackers can save the checkpoint
   - we need some way of deduping between the different checkpoints for different rollups
   - we need to make sure deposits are only redeemable on one chain
   - we probably want to include the source chain inside the signal
- Obviously this system is not designed to handle multiple L2s, but I'm not sure why.
- As far as I can tell, the deposit `id` is signalled like any other signal. This means that I can just call `sendSignal` with a valid deposit ID (without sending any ETH), and then redeem it on the other chain

Any clarifications would be appreciated

nikeshnazareth:
I have some other suggestions for things to improve. I'm happy to make the changes but I thought I'd mention it here so the rationale is clear.

- In `LibSignal`, we could reorder the functions so the `value` is the first parameter and we can treat everything as operations on the `value`. At the moment we use the syntax `value.signal()` but not `value.signaled(sender)` or `value.verifySignal(...)`. I think if we do the first then we should do the others too.
- `isSignalStored` has a comment that it returns false if the signal itself is zero. It also seems a bit weird that we save `value` as the storage value, but just check that it's non-zero when seeing if the signal exists. Should we check that it actually has the expected value for robustness? If so, we could save the hash of the value so that `0` is still a valid signal, or we could just save a specific non-zero constant for all signals.
- The `SignalService._verifySignal` function calls one of two different versions of `LibSignal.verifySignal`, depending on whether there is an account proof. I think it would be clearer to move that branch into the `LibSignal.verifySignal` function itself. I think that also fixes some naming issues (where `root` is described as a state root but it's a storage root, and there is a `stateProof` instead of a `storageProof`)
- Is the parameter `gas: gasleft()` redundant (in `_sendETH`)?

nikeshnazareth:
> The reason we do this is store store 'historical' checkpoints...

I agree with the rationale for storing historical checkpoints - I'm objecting to duplicating it. I'll create a draft for moving the tracker storage into the commitment store. I think this would also naturally let rollups overwrite obsolete checkpoints.

> For now we have decided that the SS will only allow communication between two chains L1 <-> L2.

This doesn't really make sense to me because we're treating it as a singleton and hardcoding the address. So this decision implicitly means that there can only ever be one rollup using this stack. I can understand making a simple version with core features and slowly expanding that over time (which we've already done several times in this repo) but I don't think that applies in this case because:
- we'd have to modify the interface and the processing logic once we introduce new chains.
- this version includes new functionality (eg. an authorizer address) that probably will be removed when we generalise.
I think generalising it from the start will be simpler, and will also help clarify the design / assumptions. I'll create a draft for your review.


> Maybe i dont understand the exploit, as the storage proof will only be valid against the trusted storageRoot of the signal service? You can make a valid storage proof for another account but it wont match the trusted root the signal service gets correct?

I think I'm saying something closer to this:
- imagine we have a checkpoint that includes the full state root
- someone tries to verify a proof that has an empty account proof
- the code will then incorrectly interpret the state root as a storage root (the root of a simple Merkle tree I think) and then see if the signal can be interpreted as a valid leaf (or intermediate node) in that tree
- whether it is possible for that proof to succeed depends on deeply understanding the Merkle-Patricia trie structure and under what conditions we can make it partially behave like a storage root. Since they are both variations of a merkle tree, it's not obvious to me that this is impossible (although I guess it is). Note that at this point we are not using the signal service `account`, so we can potentially focus on a more controllable part of the tree.
- but either way, the high level point is just that it makes me nervous that there is an independent path you can use to attempt an invalid proof. Can you remind me why were are supporting both cases?

LeoPatOZ:
> * In `LibSignal`, we could reorder the functions so the `value` is the first parameter and we can treat everything as operations on the `value`. At the moment we use the syntax `value.signal()` but not `value.signaled(sender)` or `value.verifySignal(...)`. I think if we do the first then we should do the others too.

This is now fixed
> * `isSignalStored` has a comment that it returns false if the signal itself is zero. It also seems a bit weird that we save `value` as the storage value, but just check that it's non-zero when seeing if the signal exists. Should we check that it actually has the expected value for robustness? If so, we could save the hash of the value so that `0` is still a valid signal, or we could just save a specific non-zero constant for all signals.

Okay i went with the `hash the underlying value approach` i think its better this way - less edge cases

> * The `SignalService._verifySignal` function calls one of two different versions of `LibSignal.verifySignal`, depending on whether there is an account proof. I think it would be clearer to move that branch into the `LibSignal.verifySignal` function itself. I think that also fixes some naming issues (where `root` is described as a state root but it's a storage root, and there is a `stateProof` instead of a `storageProof`)

Okay they are now merged

LeoPatOZ:
> * As far as I can tell, the deposit `id` is signalled like any other signal. This means that I can just call `sendSignal` with a valid deposit ID (without sending any ETH), and then redeem it on the other chain

Okay i changed this now to have a separate namespace param - one for eth deposits and one for generic signals. (Im not convinced on all the namings so open to suggestions) but at least this isnt an issue

nikeshnazareth:
I just realised I updated `LibTrieProof` but it has Taiko as a security contact. Are we treating it like an external library that needs to remain identical? If not, should we push the whole "if there is an account proof get the storage root first" logic to the `LibTrieProof` to avoid duplicating all the parameters in `LibSignal`?

ggonzalez94 (approved):
Did a final pass and I think this is ready to merged. We might want to exclude the fork tests(or have a separate command to run them), since they block the pipeline and make every forge test run fail.
I know we are changing the testing strategy for the SS and the bridge on #119 , so I'm fine with either approach as long as we keep main clean

Issue #82, opened 2025-03-21
Simplify when the existing prover wins the auction for the next period


ggonzalez94:
This shouldn't happen often, so it is not a big priority IMO, but would still be a good mechanism to have.

Right now, if you are the prover for the current period and you are outbid the logic to kickstart the next period is started, even if you go back and beat the best bid.
This works fine from the correctness perspective, but has a few drawbacks:
- You have to deposit a `livenessBond` again, since your current bond is locked as stake (less capital efficient, especially if the bond is big)
- You have to go and finalize the period, which is less eficient in terms of proving costs

Ideally, if possible we should just extend the current prover's period instead of starting  new one.

xiaodino:
It sounds like a nice optimization! I have a few questions to better understand the motivation and its effects, especially as I'm interested in contributing to implementing it. 😊

To summarize my current understanding:
- When the kickstart logic for the next proving period runs, we check if the newly selected prover is the same as the previous prover. If it is, we "just extend the current prover's period" and reuse the existing liveness bond rather than requiring a new deposit, saving gas and improving capital efficiency.
- To ensure proper economic incentives, the new bid’s `offeredFee` must always be higher than the prover's previous bid. This prevents the prover from intentionally increasing their bid repeatedly to delay proof submission and maximize profits. We want to incentivize prompt proof submission, not encourage delays.

I’m curious about your thoughts on this approach—especially regarding incentives and potential edge cases we might need to consider.

PR #91, opened 2025-03-30
Simplify when the existing prover wins the auction for the next period #82

xiaodino:
Please see https://github.com/OpenZeppelin/minimal-rollup/issues/82 for the discussion.

Any feedback is welcome

### DIFF
--- a/src/protocol/taiko_alethia/ProverManager.sol
+++ b/src/protocol/taiko_alethia/ProverManager.sol
@@ -68,7 +68,7 @@ contract ProverManager is IProposerFees, IProverManager {
 
     event Deposit(address indexed user, uint256 amount);
     event Withdrawal(address indexed user, uint256 amount);
-    event ProverOffer(address indexed proposer, uint256 period, uint256 fee, uint256 stake);
+    event ProverOffer(address indexed proposer, uint256 period, uint256 fee, uint256 stake, bool bondReused);
     event ProverEvicted(address indexed prover, address indexed evictor, uint256 periodEnd, uint256 livenessBond);
     event ProverExited(address indexed prover, uint256 periodEnd, uint256 provingDeadline);
     event NewPeriod(uint256 period);
@@ -149,6 +149,7 @@ contract ProverManager is IProposerFees, IProverManager {
     /// @dev The current best price may be the current prover's fee or the fee of the next bid, depending on whether the
     /// period is active or not.
     /// An active period is one that doesn't have an `end` timestamp yet.
+    /// When a period is active, the bid will be for the next period.
     function bid(uint256 offeredFee) external {
         uint256 currentPeriod = currentPeriodId;
         Period storage _currentPeriod = _periods[currentPeriod];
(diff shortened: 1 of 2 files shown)


src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -317,11 +325,25 @@ contract ProverManager is IProposerFees, IProverManager {
...
+        private
+        returns (bool bondReused)
+    {
         period.prover = prover;
         period.fee = fee;
-        period.stake = stake; // overwrite previous value. We assume the previous value is zero or already returned
-        balances[prover] -= stake;
+
+        // Reuse bond if same prover is continuing from previous period
+        if (_periods[currentPeriodId].prover == prover) {
```
LeoPatOZ, line 336: Not exactly sure if this will save some gas / early optimisation, but given anytime we call this function we have previously read the currentPeriod, maybe this function can just accept currentPeriod as a param avoiding
these two reads:
`_periods[currentPeriodId]`

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -317,11 +325,25 @@ contract ProverManager is IProposerFees, IProverManager {
     /// @param prover The address of the prover
     /// @param fee The fee offered by the prover
     /// @param stake The liveness bond to be staked
-    function _updatePeriod(Period storage period, address prover, uint256 fee, uint256 stake) private {
+    function _updatePeriod(Period storage period, address prover, uint256 fee, uint256 stake)
+        private
+        returns (bool bondReused)
+    {
```
LeoPatOZ, line 331:
```suggestion
    function _updatePeriod(Period storage nextPeriod, Period storage currentPeriod, address prover, uint256 fee, uint256 stake)
        private
        returns (bool bondReused)
    {
```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -317,11 +325,25 @@ contract ProverManager is IProposerFees, IProverManager {
     /// @param prover The address of the prover
     /// @param fee The fee offered by the prover
     /// @param stake The liveness bond to be staked
-    function _updatePeriod(Period storage period, address prover, uint256 fee, uint256 stake) private {
+    function _updatePeriod(Period storage period, address prover, uint256 fee, uint256 stake)
```
LeoPatOZ, line 328: nit: maybe this function also doesnt need to return if the bond is reused anymore since thats info only really useful for bidding? So the bid function could be:

```diff
    function bid(uint256 offeredFee) external {
        uint256 currentPeriod = currentPeriodId;
        Period storage _currentPeriod = _periods[currentPeriod];
        Period storage _nextPeriod = _periods[currentPeriod + 1];
+       bool bondReused;
        if (_currentPeriod.end == 0) {
            _ensureSufficientUnderbid(_currentPeriod.fee, offeredFee);
            _closePeriod(_currentPeriod, successionDelay, provingWindow);
        } else {
            address _nextProverAddress = _nextPeriod.prover;
            if (_nextProverAddress != address(0)) {
                _ensureSufficientUnderbid(_nextPeriod.fee, offeredFee);

                // Refund the liveness bond to the previous next bidder,
                // but only if it's not the current prover (who may reuse the bond)
                if (_nextProverAddress != _currentPeriod.prover) {
                    balances[_nextProverAddress] += _nextPeriod.stake;
+                    bondReused = true;
                }
            }
        }

        // Record the next period info
        uint256 _livenessBond = livenessBond;
-        _updatePeriod(_nextPeriod, msg.sender, offeredFee, _livenessBond);

        emit ProverOffer(msg.sender, currentPeriod + 1, offeredFee, _livenessBond, bondReused);
    }

```

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -317,11 +325,25 @@ contract ProverManager is IProposerFees, IProverManager {
...
+            // Carry over existing stake — no deduction needed
+            period.stake = _periods[currentPeriodId].stake;
+            return true;
+        } else {
+            // New prover or fresh start — deduct stake from balance
+            require(balances[prover] >= stake, "Insufficient balance for bond");
+            balances[prover] -= stake;
+            period.stake = stake;
+            return false;
+        }
```
LeoPatOZ, line 346: nit:
```suggestion
bondReused = _periods[currentPeriodId].prover == prover
        if (bondReused) {
            // Carry over existing stake — no deduction needed
            period.stake = _periods[currentPeriodId].stake;
        } else {
            // New prover or fresh start — deduct stake from balance
            require(balances[prover] >= stake, "Insufficient balance for bond");
            balances[prover] -= stake;
            period.stake = stake;
        }
```

src/protocol/taiko_alethia/ProverManager.sol
xiaodino: Thanks. Updated.

src/protocol/taiko_alethia/ProverManager.sol
xiaodino: Updated.

src/protocol/taiko_alethia/ProverManager.sol
xiaodino: Updated.

LeoPatOZ:
No longer going with this approach

Issue #84, opened 2025-03-24
Make DelayedInclusionStore an abstract contract


LeoPatOZ:
Related to @ernestognw PR https://github.com/OpenZeppelin/minimal-rollup/pull/75. 
We could make the DelayedInclusionStore an abstract contract then making the inbox inherit this contract.

Changes required:
1. Change `processDueInclusions` and `publishDelayed ` to public virtual functions
2. Removing constructor params by making all variables public virtual ones i.e:

convert `public immutable blobRefRegistry` into:
```solidity
function blobRefRegistry() public virtual returns (IBlobRegistry) {
  return IBlobRegistry(<CANONICAL REGISTRY>);
}
```
3. Possibly extracting the binary search algo into its own internal virtual functions (can easily be overridden if the inclusions are held in a mapping)

---
This can be a separate PR, i believe we should merge https://github.com/OpenZeppelin/minimal-rollup/pull/74 first.

ggonzalez94:
Closing this since the PR is already merged. @LeoPatOZ feel free to reopen if there was anything missing for this

PR #92, opened 2025-04-02
Abstract delayed inclusion

LeoPatOZ:
Made the delayed inclusion store an abstract contract that the inbox implements. Solves https://github.com/OpenZeppelin/minimal-rollup/issues/84


src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -31,33 +31,28 @@ contract DelayedInclusionStore is IDelayedInclusionStore {
...
-        (bytes32 refHash,) = blobRefRegistry.registerRef(blobIndices);
+    function publishDelayed(uint256[] memory blobIndices) external virtual {
+        bytes32 refHash = _getRefHash(blobIndices);
         DueInclusion memory dueInclusion = DueInclusion(refHash, block.timestamp + inclusionDelay);
         _delayedInclusions.push(dueInclusion);
         emit DelayedInclusionStored(msg.sender, dueInclusion);
     }
 
-    /// @inheritdoc IDelayedInclusionStore
+    function _getRefHash(uint256[] memory blobIndices) internal virtual returns (bytes32);
```
ggonzalez94, line 48: I'm a bit concerned that if we make this function virtual and allow implementer contracts to decide how to implement someone could use something different than the blobRefRegistry and not emit the blob number and hash in an event(going back to the discussion we had in slack).
Maybe it is worth to specifying what the implementation has to emit at least?

src/protocol/taiko_alethia/DelayedInclusionStore.sol
```diff
@@ -31,33 +31,28 @@ contract DelayedInclusionStore is IDelayedInclusionStore {
...
+
+    /// @notice Returns a list of publications that should be processed by the Inbox
     /// @dev Only returns inclusions if the delay period has passed
     /// otherwise returns an empty array.
     /// @dev Can only be called by the inbox contract.
-    function processDueInclusions() external returns (Inclusion[] memory) {
+    function processDueInclusions() public virtual returns (Inclusion[] memory) {
```
ggonzalez94, line 54: Now that this is an abstract contract this can be internal instead of public if you think about it. This is not intended to be called by anyone other than the inbox(which now implements this contract). If you do that then you can remove the require, which didn't make much sense as it was checking that this was being called by the same contract(which is equivalent of making this internal)

ggonzalez94: You would also remove this from the `IDelayedInclusionStore` interface

src/protocol/taiko_alethia/DelayedInclusionStore.sol
ggonzalez94: Actually you can just make this just an internal or private function **implemented** inside this abstract contract, and hold the reference to the BlobrefRegistry here in the `DelayedInclusionStore`. Since the Inbox inherits from the abstract contract it will also have a reference to the `blobRefRegistry`. I think this is the cleanest.
``` solidity
constructor(uint256 _inclusionDelay, address _blobRefRegistry) {
        inclusionDelay = _inclusionDelay;
        blobRefRegistry = IBlobRefRegistry(_blobRefRegistry);
    }
```

test/DelayedInclusionStore.t.sol
ggonzalez94, line 1: Maybe it makes more sense to make this the test file for the Inbox contract and later we add tests for the rest of the functions? Since you are already using the TaikoInbox itself I believe that makes more sense. The other option is to make these tests just use a minimal contract that inherits from the  `DelayedInclusionStore`, but I'm not sure this is worth it.

src/protocol/IDelayedInclusionStore.sol
```diff
@@ -10,6 +10,9 @@ interface IDelayedInclusionStore {
     /// @param inclusionsList list of inclusions
     event DelayedInclusionProcessed(Inclusion[] inclusionsList);
 
+    // @dev When caller is not the inbox contract
+    error OnlyInbox();
```
ggonzalez94, line 14: we should be able to remove this error if we make `processDueInclusions` internal

ggonzalez94 (commented):
Looking good. Left a few suggestions that I think are important to solve before we merge it

src/protocol/taiko_alethia/DelayedInclusionStore.sol
LeoPatOZ: yeah i just didnt know wether to put the` blobRefRegistry` variable inside the inbox or delayed inclusion. I was forcing it to be in the inbox hence the weird workaround. Will change!

src/protocol/taiko_alethia/DelayedInclusionStore.sol
LeoPatOZ: This is the annoying part ... i kept it because otherwise theres no way to specifically test this function :(

src/protocol/taiko_alethia/DelayedInclusionStore.sol
LeoPatOZ: done!

test/DelayedInclusionStore.t.sol
LeoPatOZ: Yeah i agree - is there anything specific you had in mind. For now i just made an inbox folder in tests, then moved the test their to make it more explicit that this is an 'inbox' test

src/protocol/IDelayedInclusionStore.sol
LeoPatOZ: okay fixed

test/DelayedInclusionStore.t.sol
LeoPatOZ: okay made a mock taiko inbox to expose the function

PR #97, opened 2025-04-04
Price delayed publications

### LINKED ISSUE #68: Price delayed publications for proving and modify the ProverMarket
As part of the ahead of time auction [design](https://github.com/OpenZeppelin/minimal-rollup/pull/62), we have the idea of pricing delayed publications differently.
If we go down this path, calculating rewards when there are delayed publications becomes more tricky, at least with the [current proposed implementation](https://github.com/OpenZeppelin/minimal-rollup/pull/49). This is because we are currently not storing in the `Period` struct wether a publication is delayed or not.
Once we decide if we want to price them differently, we should implement a mechanism to do so in the `ProverManager`

ggonzalez94:
Fixes #68 

This PR is an attempt to give the flexibility to price delayed publications differently, while still keeping them as part of the fees(more details can be found on the issue).

Pros of this approach:
- Flexibility, delayed publications can be priced differently than regular pubs(we are using a uint16, so they can represent up to ~6.55x the normal fee - alternatively we can use a uint24 and that would allow ~167.77x)
- Delayed fees are treated just as regular fees, and payed to provers when calling prove(no need to wait until the end of the period or put them in the stake)

Cons:
For now it:
- <del> Uses more storage since we have a new variable for the `delayedFeePercentage`. This shouldn't be an issue once we pack the storage variables</del>
- <del>Adds another config parameter(which will be abstracted with an internal function soon)</del>

These both are fixed in #100 

This PR also:
- Removes the ability to deposit from the `payPublicationFee` function
- Fixes a bug where the Inbox was always sending the publication fee as part of the call to `payPublicationFee`. Now we rely on the proposer to deposit before and the call this function. This saves some gas since we don't need to query the fees beforehand

### DIFF
--- a/src/protocol/BaseProverManager.sol
+++ b/src/protocol/BaseProverManager.sol
@@ -1,14 +1,17 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.28;
 
+import {LibPercentage} from "../libs/LibPercentage.sol";
 import {ICheckpointTracker} from "./ICheckpointTracker.sol";
 import {IProposerFees} from "./IProposerFees.sol";
 import {IProverManager} from "./IProverManager.sol";
 import {IPublicationFeed} from "./IPublicationFeed.sol";
+
 import "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
 abstract contract BaseProverManager is IProposerFees, IProverManager {
     using SafeCast for uint256;
+    using LibPercentage for uint96;
 
     struct Period {
         // SLOT 1
@@ -17,7 +20,7 @@ abstract contract BaseProverManager is IProposerFees, IProverManager {
         // SLOT 2
         // the fee that the prover is willing to charge for proving each publication
         uint96 fee;
-        // the percentage (in bps) of the fee that is charged for delayed publications.
+        // the percentage (with two decimals precision) of the fee that is charged for delayed publications.
         uint16 delayedFeePercentage;
         // the timestamp of the end of the period. Default to zero while the period is open.
         uint40 end;
@@ -94,7 +97,7 @@ abstract contract BaseProverManager is IProposerFees, IProverManager {
         // Deduct fee from proposer's balance
         uint96 fee = _periods[periodId].fee;
         if (isDelayed) {
-            fee = _calculatePercentage(fee, _periods[periodId].delayedFeePercentage).toUint96();
+            fee = fee.scaleByPercentage(_periods[periodId].delayedFeePercentage);
         }
         _balances[proposer] -= fee;
     }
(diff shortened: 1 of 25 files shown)


src/protocol/IProverManager.sol
```diff
@@ -5,6 +5,30 @@ import {ICheckpointTracker} from "./ICheckpointTracker.sol";
 import {IPublicationFeed} from "./IPublicationFeed.sol";
 
 interface IProverManager {
+    /// @notice Emitted when a prover bids to prove a period
+    /// @param prover The address of the prover that made the bid
+    /// @param period The period that the prover is bidding to prove
+    /// @param fee The fee that the prover is willing to charge for proving each publication
+    /// @param stake The stake that the prover is going to put as stake for the period
+    event ProverOffer(address indexed prover, uint256 period, uint256 fee, uint256 stake);
```
LeoPatOZ, line 13: nit: This is true for other events, we have `period` to mean periodID, since we have a `Period` struct, we should rename to `periodId`

src/protocol/taiko_alethia/ProverManager.sol
```diff
@@ -58,6 +66,10 @@ contract ProverManager is IProposerFees, IProverManager {
     /// @notice The percentage (in bps) of the remaining liveness bond rewarded to the prover who proves the final
     /// publication after the deadline
     uint256 public immutable rewardPercentage;
+    /// @notice The percentage (in bps) of the fee that is charged for delayed publications
+    /// @dev It is recommended to set this to >10,000 bps since delayed publications should usually be charged at a
+    /// higher rate
+    uint16 public immutable delayedFeePercentage;
```
LeoPatOZ, line 72: This is probably very rare but is uint16 enough? given the percentage is  given in bps. (we could only charge up to ~6.5x the normal fee)

ggonzalez94: Correct. I have a feeling that's more than enough, but we could make it an `uint24` and that would allow ~167.77x. For now it wouldn't do any impact on storage since we are only using 25 bytes, so we still have 7 bytes available in that slot. But when the protocol evolves those bytes might be useful for some other purposes.

src/protocol/BaseProverManager.sol
```diff
@@ -0,0 +1,383 @@
...
+    function withdraw(uint256 amount) external {
+        _balances[msg.sender] -= amount;
+        _transferOut(msg.sender, amount);
+        emit Withdrawal(msg.sender, amount);
+    }
+
+    /// @inheritdoc IProposerFees
+    /// @dev This function advances to the next period if the current period has ended.
+    function payPublicationFee(address proposer, bool isDelayed) external {
+        require(msg.sender == inbox, "Only the Inbox contract can call this function");
```
nikeshnazareth, line 83: There are several examples of error strings rather than custom errors (as required by the style guide). It's probably worth creating them as we go.

ggonzalez94:
Merging this since it's been outstanding for a while. If we find anything that we'd like to change we have the proving period PR still open

PR #98, opened 2025-04-04
refactor: Refactor ProverManager to adhere to coding guidelines

ggonzalez94:
Refactor variable names and conventions to adhere to the new coding guidelines

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -53,6 +53,7 @@ This guide is not intended to be comprehensive, but rather to provide a starting
 1. Events and structs when part of the contract's public API should be defined in the contract interface.
 
     If they are an implementation detail, or emitted by a specific implementation, they should instead be defined in the contract itself, and not part of the interface.
+
     ```solidity
     interface IMyContract {
         // These event and struct are part of the interface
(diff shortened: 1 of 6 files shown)


ggonzalez94:
> 1. Constructor missing natspec
I've seen protocols with and without natspec for constructors. I don't have a strong opinion, but I agree it doesn't hurt. I added some guidance to our style guide.

> _claimProvingVacancy is missing [natspec](https://github.com/OpenZeppelin/minimal-rollup/blob/main/src/protocol/taiko_alethia/ProverManager.sol#L297)
Done ✅ 

> _deposit doesnt follow [natspec](https://github.com/OpenZeppelin/minimal-rollup/blob/main/src/protocol/taiko_alethia/ProverManager.sol#L290) although i dont mind
Agree that adding `@param` just for the sake of it makes it more verbose without any real benefit. I left it as it was

> Some events can be moved to interface, maybe Deposit Withdrawal ProverOffer, as these are generic enough?
I think it makes sense. I moved some of them to `IProverManager` and the rest to `IProposerFees`, let me know what you think.

> We don't have custom errors
I was planning to do this at some point later along with other contracts that we started with simple `require` if you are ok with it

> Not sure we discussed but i prefer struct comments to be above the member name.
Yeah I'm ok with this. I deleted some redundant comments and move the others above each member

LeoPatOZ (approved):

Issue #51, opened 2025-02-26
Allow using ERC20 tokens in the ahead of time auction


ggonzalez94:
The current implementation only allows ETH. We should also allow ERC20 tokens(e.g. TAIKO) for prover incentives

ggonzalez94:
There are 3 possible approaches to this:
- Same contract, same functions
- Same contract, but different functions(this is the one I like the least)
- Base contract with common logic and two contracts that inherit from it(one for eth and other for ERC20)

Considering we are already planning to have a base implementation with common logic(#72 ) the third option seems like the best approach, and also the most gas efficient.

PR #100, opened 2025-04-07
Prover manager abstraction and support ERC20 tkens

ggonzalez94:
Fixes #92 and #51 

This PR should be merged after #97 and #98 - I'll change the target branch to main once we merge those and there are still some small things I'll work on this branch, but the main idea is there.

## Description
This PR:
- Makes the ProverManager an abstract contract with all the previous immutable parameters now internal virtual functions that can be overriden
- The ProverManager is now unaware of the currency and we have two contracts that inherit from it: `ETHProverManager` and `ERC20ProverManager`
- Packs the Period struct to use only 2 slots. I don't expect the struct to change much after this, so I thought it was a good moment to pack the storage slots already. I chose what fields to pack prioritizing the most used functions and what we care most about making cheaper(in order: `payPublicationFee`, `prove` and `bid`)
    - timestamps use `uint40` which is good until the year 35,073
    - fees and balances use `uint96` which is significantly bigger than the current ETH supply and the TKO supply(1B). We could be more agressive on this, but we won't get under 2 slots for the struct, so I think there's no need to do it
- I considered making `Period` a library as previously suggested, but since this is the only contract that uses it I thought for now it might not be worth it and will just add another level of indirection

## Some Nuances
`ERC20ProverManager` transfers funds from the `initialProver` in the constructor. This is to ensure there is a liveness bond from the start. But this means that there has to be a token approval from the initial prover to this contract(which has not been deployed yet). This is solved by deploying `ERC20ProverManager` using CREATE2. That way we can calculate the address beforehand and make the approval. Another option is to make the initial period not have a stake, but that feels less safe.

## Open Questions
- In this PR all 3 prover managers are abstract, because we don't have specified values for all configuration parameters. Should we define some sensible defaults in the `BaseProverManager`? Or do we prefer having that discussion in the future?


test/ERC20ProverManager.t.sol
```diff
@@ -0,0 +1,279 @@
...
+    LIVENESS_WINDOW,
+    MAX_BID_PERCENTAGE,
+    PROVING_WINDOW,
+    REWARD_PERCENTAGE,
+    SUCCESSION_DELAY
+} from "./BaseProverManager.t.sol";
+
+import {IERC20Errors} from "@openzeppelin/contracts/interfaces/draft-IERC6093.sol";
+
+contract ERC20ProverManagerMock is ERC20ProverManager {
```
LeoPatOZ, line 36: put this in the mock folder?

test/ETHProverManager.t.sol
```diff
@@ -0,0 +1,155 @@
...
+    INITIAL_PERIOD,
+    LIVENESS_BOND,
+    LIVENESS_WINDOW,
+    MAX_BID_PERCENTAGE,
+    PROVING_WINDOW,
+    REWARD_PERCENTAGE,
+    SUCCESSION_DELAY
+} from "./BaseProverManager.t.sol";
+
+contract ETHProverManagerMock is ETHProverManager {
```
LeoPatOZ, line 32: in the mock folder as well?

src/protocol/ETHProverManager.sol
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    function _transferOut(address to, uint256 amount) internal override {
+        bool ok;
+        // Using assembly to avoid memory allocation costs; only the call's success matters to ensure funds are sent.
+        assembly ("memory-safe") {
+            ok := call(gas(), to, amount, 0, 0, 0, 0)
+        }
+        require(ok, "Withdraw failed");
+    }
```
LeoPatOZ, line 35: should we add:
```solidity
    receive() external payable {
_deposit(msg.sender, msg.value);
    }
```
as a fallback

src/protocol/BaseProverManager.sol
```diff
@@ -203,10 +139,11 @@ contract ProverManager is IProposerFees, IProverManager {
...
         // We use this to mark the prover as evicted
-        (uint256 end,) = _closePeriod(period, exitDelay, 0);
+        (uint40 end,) = _closePeriod(period, _exitDelay(), 0);
 
         // Reward the evictor and slash the prover
-        uint256 evictorIncentive = _calculatePercentage(period.stake, evictorIncentivePercentage);
+        // Casting this directly is safe since the resulting number is smaller than the original value
+        uint96 evictorIncentive = uint96(_calculatePercentage(period.stake, _evictorIncentivePercentage()));
         balances[msg.sender] += evictorIncentive;
         period.stake -= evictorIncentive;
```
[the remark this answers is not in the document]
ggonzalez94: They will actually be two separate values, since we are substracting `evictorIncentive` from the `period.stake`. We could use intermediate variables, but it will get uglier and I expect the compiler to be smart enough to do the optimizations(but will double check with a gas report)

src/protocol/ERC20ProverManager.sol
```diff
@@ -0,0 +1,46 @@
...
+        address _inbox,
+        address _checkpointTracker,
+        address _publicationFeed,
+        address _initialProver,
+        uint96 _initialFee,
+        address _token,
+        uint256 _initialDeposit
+    ) BaseProverManager(_inbox, _checkpointTracker, _publicationFeed, _initialProver, _initialFee, _initialDeposit) {
+        require(_token != address(0), "Token address cannot be 0");
+        require(
```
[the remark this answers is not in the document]
ggonzalez94: Usually I expect that to be the case. But this gives them the flexibility if they want to deposit a larger amount when deploying the contract(e.g. for participating in future bidding periods or if they want to also be a proposer)

test/ERC20ProverManager.t.sol
ggonzalez94: I feel it is very specific to this test(and not reusable across other tests), plus it has the imports from the `BaseProverManager.t.sol` - so I think it might be clearer to keep it under this file. Once we have default parameters(which I think the `BaseProverManager` should offer) we might be able to get rid of the mock entirely

test/ETHProverManager.t.sol
ggonzalez94: Same as above 🆙

src/protocol/BaseProverManager.sol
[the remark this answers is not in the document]
ggonzalez94: They will actually be two separate values, since we are substracting evictorIncentive from the period.stake. We could use intermediate variables, but it will get uglier and I expect the compiler to be smart enough to do the optimizations(but will double check with a gas report)

ggonzalez94:
>  I was also thinking should we add interfaces to the abstract contract like:
> 
> IErc20Receiver {
> External deposit(amount)
> }
> 
> And the same for the ETH one?

yeah, I think this is useful for anyone using the contract they can import the interfaces vs the implementation. I added the interfaces to the `IProverManager` interface since at least now they are only used there, but I don't have a strong opinion so happy to move them to their own file as well(I just didn't want to clutter the repo with more files preemptively)

LeoPatOZ (approved):
LGTM

Issue #79, opened 2025-03-20
Integrate test coverage in the CI


ggonzalez94:
Report test coverage on every PR and on the main branch.

MUST:
- Show the coverage on the CI logs
- Upload to Codecov for easy visualization
- Ignore vendor files and mocks

SHOULD:
- Show the diff for every PR so we know if coverage is improving or decreasing

PR #106, opened 2025-04-17
gas diff

LeoPatOZ:
Solves https://github.com/OpenZeppelin/minimal-rollup/issues/79 
Generate gas report diffs using this CI tool (https://github.com/Rubilmax/foundry-gas-diff).

For now it does not use codecov (i need an api key), it uses sticky comments for visualisation. 

Here is an example on a test branch of the output:

Screenshot 2025-04-22 at 11 31 36

BUG:
Unrelated but it seems that runnig forge test --gas-report makes some test fail while regular foundry test passes.


ggonzalez94:
> BUG:
> Unrelated but it seems that runnig forge test --gas-report makes some test fail while regular foundry test passes.

This was annoying me so I spent some time debugging, and after hitting my head agains the wall for a while I believe it is a bug in Foundry. Whenever you use `vm.blobhashes()` in any of the test it fails when running with `--gas-report` flag even if they succeed without it. I opened an issue in [their repo](https://github.com/foundry-rs/foundry/issues/10365)

gasreport.ansi
ggonzalez94, line 1: I don't think this is supposed to be committed right?

ggonzalez94 (commented):
@LeoPatOZ  Looks great, just the `ansi` file shouldn't be committed to the repo I believe.

I did a small change in the CI so that it only runs when relevant files are changed and it ignores negative tests(the ones that revert) and fuzz tests. Please take a look

gasreport.ansi
LeoPatOZ: I believe it is - thats what the action compares against.  https://github.com/Rubilmax/foundry-gas-diff/blob/main/README.md?plain=1#L71-L73

LeoPatOZ: yes you are right i tested it - we dont need the gas report. Updated it now

ggonzalez94 (approved):
LGTM!

PR #107, opened 2025-04-21
Generalise Signal Service

nikeshnazareth:
As noted on #89 (see [here](https://github.com/OpenZeppelin/minimal-rollup/pull/89#issuecomment-2800093911)), I think it will be simpler to create a multi-rollup signal service right from the start. This is WIP but since I will be unavailable this week I thought I should push the current version for your consideration.

The overall intention is:
- [x] Avoid duplicate checkpoints. The TaikoAnchor and CheckpointTracker now use the signal service for their checkpoint/l1blocks storage.
- [ ] Remove commitment access control. Anyone can post commitments, which are saved under the source address. Access control can be achieved on the recipient end (they should only use checkpoints saved by a valid address). This makes it possible to support multiple rollups. At this point I have made the change to the CommitmentStore and propagated them to the TaikoAnchor and CheckpointTracker, but not to the rest of the files. The codebase is currently inconsistent
- [ ] Generalise the ETH bridge to handle multiple rollups. This is not implemented. I suspect we will want to record which chain holds which deposits (so a malicious rollup cannot withdraw funds held by another rollup). To keep it symmetric, the L2 side of the bridge will be deployed with infinite ETH held by the L1.
- [ ] Allow the CheckpointTracker genesis to start from an arbitrary publication ID (or perhaps the current publication ID). This makes sense as long as we have a shared publication feed, because a new rollup should not have to process all the historical publications before it was deployed.

### DIFF
--- a/src/protocol/CheckpointTracker.sol
+++ b/src/protocol/CheckpointTracker.sol
@@ -7,14 +7,10 @@ import {IPublicationFeed} from "./IPublicationFeed.sol";
 import {IVerifier} from "./IVerifier.sol";
 
 contract CheckpointTracker is ICheckpointTracker {
-    /// @notice The current proven checkpoint representing the latest verified state of the rollup
-    /// @dev Previous checkpoints are not stored here but are synchronized to the `SignalService`
+    /// @notice The publication id of the current proven checkpoint representing the latest verified state of the rollup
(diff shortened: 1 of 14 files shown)


LeoPatOZ:
Closing as we have integrated the ideas  into  https://github.com/OpenZeppelin/minimal-rollup/pull/89

PR #111, opened 2025-04-25
Removes start publication from checkpoint

LeoPatOZ:
As we have decided not to allow for parallel proving (i.e. proving publications not starting from the latest proven), we no longer should / need to accept 'start' checkpoint when proving a transition - it will always be the latest proven checkpoint.

Slightly more gas efficient

Im not sure if we need to check L238 in the [prover manager](https://github.com/OpenZeppelin/minimal-rollup/blob/main/src/protocol/taiko_alethia/ProverManager.sol#L238)

### DIFF
--- a/src/protocol/CheckpointTracker.sol
+++ b/src/protocol/CheckpointTracker.sol
@@ -35,31 +35,24 @@ contract CheckpointTracker is ICheckpointTracker {
     }
 
     /// @inheritdoc ICheckpointTracker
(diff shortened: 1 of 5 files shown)


ggonzalez94:
@LeoPatOZ I made the changes we discussed yesterday so we can merge this PR. Please take a quick look

ggonzalez94 (approved):

PR #113, opened 2025-04-30
[2/2] Decouple ETH Bridge from signal service

LeoPatOZ:
PR solves:
- [x] Generalise the ETH bridge to handle multiple rollups.


src/protocol/ETHBridge.sol
```diff
@@ -12,41 +13,44 @@ contract ETHBridge is IETHBridge {
...
-    }
-
-    /// @dev Processes deposit claim by id.
-    /// @param id Identifier of the deposit
-    /// @param ethDeposit Deposit to process
-    function _processClaimDepositWithId(bytes32 id, ETHDeposit memory ethDeposit) internal virtual {
+    // TODO: Non reentrant
+    function claimDeposit(ETHDeposit memory ethDeposit, uint256 height, bytes memory proof) external {
+        bytes32 id = _generateId(ethDeposit);
+        signalService.verifySignal(height, trustedCommitmentPublisher, ethDeposit.from, id, proof);
```
nikeshnazareth, line 53: This is a bug. The `sender` should be the source `ETHBridge` contract, not `ethDeposit.from`

src/protocol/ETHBridge.sol
```diff
@@ -12,41 +13,50 @@ contract ETHBridge is IETHBridge {
...
-    {
-        id = _generateId(ethDeposit);
-    }
-
-    /// @dev Processes deposit claim by id.
-    /// @param id Identifier of the deposit
-    /// @param ethDeposit Deposit to process
-    function _processClaimDepositWithId(bytes32 id, ETHDeposit memory ethDeposit) internal virtual {
+    // TODO: Non reentrant
+    function claimDeposit(ETHDeposit memory ethDeposit, uint256 height, bytes memory proof) external {
```
ggonzalez94: May I suggest a slight reordering and spacing in between just for clarity?

```solidity
function claimDeposit(ETHDeposit memory ethDeposit, uint256 height, bytes memory proof) external {
        bytes32 id = _generateId(ethDeposit);
        require(!claimed(id), AlreadyClaimed());

        signalService.verifySignal(height, trustedCommitmentPublisher, counterpart, id, proof);
        
        _claimed[id] = true;
        _sendETH(ethDeposit.to, ethDeposit.amount, ethDeposit.data);
        emit DepositClaimed(id, ethDeposit);
    }
```

ggonzalez94 (commented):
Looks great! My only big question is if we should use the same address for the EthBridge on both chains or keep the `counterpart` variable.
I also complete the TODO by making `claimDeposit` nonReentrant using transient storage

src/protocol/ETHBridge.sol
LeoPatOZ: done

ggonzalez94 (approved):
LGTM! I haven't reviewed the tests, but I think we can start merging these PRs and I'll take a look at tests before we merge the big one into main

PR #114, opened 2025-04-30
Use two decimals for delayed scale factor

nikeshnazareth:
This PR make two changes.

Firstly, move the percentage calculations to a separate library. This is cleaner in code and it also makes it easier to use different precision amounts.

The second is to use two decimals of precision for the delayed fee percentage. As a uint16 there are 65536 possible values but in basis points we can only represent numbers between 0 and 6.5 (albeit with very high precision). With two decimals we can represent numbers up to 655 (with less precision).

### DIFF
--- a/src/libs/LibPercentage.sol
+++ b/src/libs/LibPercentage.sol
@@ -7,7 +7,7 @@ library LibPercentage {
     using SafeCast for uint256;
 
     // COMMON PRECISION AMOUNTS (https://muens.io/solidity-percentages)
-    uint256 constant BASIS_POINTS = 10000;
+    uint256 constant BASIS_POINTS = 10_000;
     uint256 constant PERCENT = 100;
 
     /// @dev Calculates the percentage of a given value scaling by `precision` to limit rounding loss
--- a/src/libs/LibPercentage.sol
+++ b/src/libs/LibPercentage.sol
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
+
+library LibPercentage {
+    using SafeCast for uint256;
+
+    // COMMON PRECISION AMOUNTS (https://muens.io/solidity-percentages)
+    uint256 constant BASIS_POINTS = 10000;
+    uint256 constant TWO_DECIMALS = 100;
+
+    /// @dev Calculates the percentage of a given value scaling by `precision` to limit rounding loss
+    /// @param value The number to scale
+    /// @param percentage The percentage expressed in `precision` units.
+    /// @param precision The precision of `percentage` (e.g. percentage 5000 with BASIS_POINTS precision is 50%).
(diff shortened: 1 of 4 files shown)


src/libs/LibPercentage.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
+
+library LibPercentage {
+    using SafeCast for uint256;
+
+    // COMMON PRECISION AMOUNTS (https://muens.io/solidity-percentages)
+    uint256 constant BASIS_POINTS = 10000;
```
LeoPatOZ, line 10:
```suggestion
    uint256 constant BASIS_POINTS = 10_000;
```

LeoPatOZ (approved):

src/libs/LibPercentage.sol
```diff
@@ -0,0 +1,29 @@
...
+    /// @return _ The scaled value
+    function scaleBy(uint256 value, uint16 percentage, uint256 precision) internal pure returns (uint96) {
+        return (value * percentage / precision).toUint96();
+    }
+
+    /// @dev Calculates the percentage (represented in basis points) of a given value
+    /// @param value The number to scale
+    /// @param percentage The percentage expressed in basis points
+    /// @return _ The scaled value
+    function scaleBy(uint256 value, uint16 percentage) internal pure returns (uint96) {
```
ggonzalez94, line 26: I'm hesitant to name the two functions the same. it might be easy to get confused by anyone using the library.
What do you think about this approach?
- `scaleBy` remains the common function that takes a `precision`
- `scaleByBps` is the bps implementation
- `scaleByPercentage` is the percentage implementation

It is more verbose, but I believe it is less error prone

test/BaseProverManager.t.sol
```diff
@@ -974,13 +974,18 @@ abstract contract BaseProverManagerTest is Test {
...
-    function _calculatePercentage(uint256 amount, uint16 percentage) internal pure returns (uint256) {
+    function _calculatePercentageBPS(uint256 amount, uint16 percentage) internal pure returns (uint256) {
         return amount * percentage / 10_000;
     }
 
+    function _calculatePercentage(uint256 amount, uint16 percentage) internal pure returns (uint256) {
+        return amount * percentage / 100;
+    }
+
+
```
ggonzalez94, line 988: Should the tests just use the library? I don't think we are trying to check if the library works properly on these tests(we can create a new unit test for `LibPercentage` if we wanted to do this)

ggonzalez94 (commented):
Looks good! Just a couple of comments

ggonzalez94 (approved):
Looks great!

ggonzalez94:
Merging this now that it has 2 approves. I had to change the foundry version that is used in the CI to stable because nightly was causing a couple tests that use `blobhash` to fail. This is likely more an issue on Foundry's side than ours, but still good to keep an eye

PR #115, opened 2025-04-30
Proving Period Library

nikeshnazareth:
The PR is almost a pure refactor. The only functional changes are:
- a bug fix in `finalizePastPeriod` to ensure the target period is closed.
- `finalizePastPeriod` now zeroes out the prover and requires the period to have a non-zero prover. This means it can only be called once per period.

The main changes are moving period-related functionality to a new library for better readability and encapsulation. It also moves the balance accounting and configuration parameters into their own abstract contracts for better encapsulation.

### DIFF
--- a/src/libs/LibProvingPeriod.sol
+++ b/src/libs/LibProvingPeriod.sol
@@ -4,7 +4,6 @@ pragma solidity ^0.8.28;
 import {LibPercentage} from "../libs/LibPercentage.sol";
 import "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
-
 /// Each proving period is defined by
 /// - the _prover_ address that must post proofs for any publications received during this period
 /// - the standard proving _fee_ that the prover charges per publication in this period
@@ -38,6 +37,18 @@ library LibProvingPeriod {
         uint40 deadline;
         // whether the final proof came after the deadline
         bool pastDeadline;
+    }
+
+    /// @notice Initializes the period with the given parameters.
+    /// @dev The _end_ and _deadline_ default to zero. The _pastDeadline` flag defaults to false.
+    /// @dev This can be called multiple times to set the latest bid while the auction is ongoing.
+    function init(Period storage period, address prover, uint96 fee, uint16 delayedFeePercentage, uint96 stake)
+        internal
+    {
+        period.prover = prover;
+        period.fee = fee;
+        period.delayedFeePercentage = delayedFeePercentage;
+        period.stake = stake;
     }
 
     /// @notice The period has an end timestamp in the past
--- a/src/libs/LibProvingPeriod.sol
+++ b/src/libs/LibProvingPeriod.sol
@@ -87,6 +87,10 @@ library LibProvingPeriod {
         return block.timestamp > period.deadline && period.deadline != 0;
     }
 
+    function isVacant(Period storage period) internal view returns (bool) {
+        return !isInitialized(period) && isOpen(period);
+    }
+
     function totalFeeEarned(Period storage period, uint256 numPublications, uint256 numDelayedPublications)
         internal
         view
--- a/src/libs/LibProvingPeriod.sol
+++ b/src/libs/LibProvingPeriod.sol
@@ -45,6 +45,7 @@ library LibProvingPeriod {
     function init(Period storage period, address prover, uint96 fee, uint16 delayedFeePercentage, uint96 stake)
         internal
     {
+        require(prover != address(0), "Prover cannot be zero address");
         period.prover = prover;
         period.fee = fee;
         period.delayedFeePercentage = delayedFeePercentage;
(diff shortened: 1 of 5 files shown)


src/libs/LibProvingPeriod.sol
```diff
@@ -0,0 +1,141 @@
...
+        return isDelayed ? period.fee.scaleBy(period.delayedFeePercentage, LibPercentage.PERCENT) : period.fee;
+    }
+
+    /// @notice The period has no end timestamp
+    function isOpen(Period storage period) internal view returns (bool) {
+        return period.end == 0;
+    }
+
+    /// @notice The timestamp is after the end of the period (which must be set)
+    function isBefore(Period storage period, uint256 timestamp) internal view returns (bool) {
```
ggonzalez94, line 76: I believe the name of this function is a bit confusing. May I suggest this instead? Since that better reflects that the timestamp you are checking for is after the end of the period
```suggestion
    function isAfterEnd(Period storage period, uint256 timestamp) internal view returns (bool) {
```

src/libs/LibProvingPeriod.sol
```diff
@@ -0,0 +1,141 @@
...
+        return period.end != 0 && timestamp > period.end;
+    }
+
+    /// @notice The timestamp is not after the end of the period
+    function isNotBefore(Period storage period, uint256 timestamp) internal view returns (bool) {
```
ggonzalez94, line 81: same here
```suggestion
    function isNotAfterEnd(Period storage period, uint256 timestamp) internal view returns (bool) {
```

Another option is isBeforeOrAtEnd

ggonzalez94:
> Before this commit the finalizePastPeriod function confirmed there
> is a publication after period.end but it did not confirm that
> period.end was non-zero

Good catch. This only happen because we have a shared publication feed right? Once we move to having the publications per inbox, we shouldn't have to worry about this?

src/protocol/ProverManagerConfig.sol
```diff
@@ -0,0 +1,42 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+abstract contract ProverManagerConfig {
+    /// @dev Returns the maximum percentage (in bps) of the previous bid a prover can offer and still have a successful
+    /// bid
+    /// @return _ The maximum bid percentage value
+    function _maxBidPercentage() internal view virtual returns (uint16);
```
ggonzalez94, line 8: Now that we express fractions of values both as bps and percentage perhaps we should rethink how we call these functions?

ggonzalez94 (commented):
Looks great! Just left a few comments, most are nits to improve clarity

LeoPatOZ (approved):

src/libs/LibProvingPeriod.sol
nikeshnazareth: I think it makes sense to write it in terms of the period, because that's show it shows up when the library is used. Eg. consider this existing line:

`require(period.isBefore(provenPublication.timestamp), PublicationNotAfterPeriod());`

vs the suggested:

`require(period.isAfterEnd(provenPublication.timestamp), PublicationNotAfterPeriod());`

src/libs/LibProvingPeriod.sol
nikeshnazareth: Same comment as above. Consider the existing

`require(period.isNotBefore(lastPub.timestamp), LastPublicationIsAfterPeriod());`

vs the suggested

`require(period.isNotAfterEnd(lastPub.timestamp), LastPublicationIsAfterPeriod());`

src/libs/LibPercentage.sol
```diff
@@ -32,6 +32,6 @@ library LibPercentage {
     /// @param percentage The percentage to single-percentage precision (e.g. percentage 50 is 50%)
     /// @return _ The scaled value
     function scaleByPercentage(uint256 value, uint16 percentage) internal pure returns (uint96) {
-        return scaleBy(value, percentage, BASIS_POINTS);
+        return scaleBy(value, percentage, PERCENT);
```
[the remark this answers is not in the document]
nikeshnazareth: Good catch, thanks!

nikeshnazareth:
> > Before this commit the finalizePastPeriod function confirmed there
> > is a publication after period.end but it did not confirm that
> > period.end was non-zero
> 
> Good catch. This only happen because we have a shared publication feed right? Once we move to having the publications per inbox, we shouldn't have to worry about this?

I think that's unrelated. This bug was saying you can finalize an Open period. We want to finalize periods that have been completely proven, and we do this by ensuring there is a proven publication that occurs after the `end` of the period, which implies the last publication in the period is also proven. But if `end` is 0 then every proven period passes this check. That's true whether or not we have shared publications.

src/protocol/ProverManagerConfig.sol
nikeshnazareth: Good point. I changed all basis point values to "fraction" rather than "percentage". Note that I left `_delayedFeePercentage` because that really is a percentage

ggonzalez94 (approved):
LGTM! Fixed some merge conflicts and merging this to main now

PR #116, opened 2025-05-01
Message Relayer

LeoPatOZ:
In order to abstract away fees / gas limits, we can introduce a third party contract that relays the deposit message and deals with fees.  The simple flow is:
- User specifies the deposit is to be fulfilled by a relayer by setting the 'to' address to a relayer contract (of their choosing)
- Relayer contract calls `claimDeposit` on behalf of user - receiving the deposited ETH
- It forwards the call to any contract / EOA as specified by the user 
- Takes a fee (specified by the user in the data) for completing this task

The relayer profits the difference between the fee and any gas cost incurred by calling `relayMessage`. It is up to the relayer to determine if this is a profitable trade for them. The user will always get a determined amount (i.e amount deposited - fee).

I suspect other protocols / relayers might deploy their own relayer contract with a more complex fee mechanism.


nikeshnazareth:
I made some minor changes. Looks good except:

- I think this confuses the "relayer" (the address that calls `relayMessage`) with the "fee recipient" (the address saved in the RELAYER slot and is sent the fees at the end). 
- I'm not a fan of "fee" as a name because it's chosen by the depositor. Perhaps "relayIncentive" or "tip" would be better. But I won't insist because it's common to use "fee" in this situation.
- `test_claimDeposit` still fails (although that is unrelated to this PR).

I'll leave it up to you to decide which of these comments to accept or ignore.

nikeshnazareth (approved):

nikeshnazareth:
> Do we need to worry about EIP-150, I've seen multiple bridges implement some gas forwarding check even taiko https://github.com/taikoxyz/taiko-mono/blob/main/packages/protocol/contracts/shared/bridge/Bridge.sol#L644. Maybe its not an issue here im not very familiar with the vulnerability

Good question. I think it's less relevant to us but the check is still advisable. My rationale:
- As it stands, we allow the caller to specify a `gasLimit` that should be provided with the call to the recipient
- Most of the time, this is not necessary, because our relay explicitly reverts if the call fails for any reason (including out-of-gas)
- This would only be a problem if it were possible for the relayer to choose an overall transaction limit such that `gasleft()` was unexpectedly high or low, which caused the recipient call to behave incorrectly and still succeed. This would be an unusual scenario, but since this is a generic relayer, we should account for that possibility.
- Our `gasLimit <= gasleft()` now (almost) ensures the relayer provides enough gas, and the `gas: gasLimit` parameter to the recipient call ensures the relayer cannot provide too much gas (or if they do, it is not forwarded to the external call).
- However, because of EIP-150, the first check is not precise. It is possible for the relayer to provide enough gas such that `gasleft` is over the `gasLimit`, but 63 / 64 * `gasleft` (which is the maximum amount actually sent to the recipient) is under. If the recipient behaves meaningfully different  (without running out of gas) within that window, then there is room for shenanigans.
- As mentioned, this would be an unusual scenario, but I think we should account for it

In terms of how to implement it, I think the before-call check is more correct (but also more fragile because it depends on opcode pricing). See [this blog](https://ronan.eth.limo/blog/ethereum-gas-dangers/) particularly the "Workaround Against “Insuficient Gas Griefing attack” for an explanation of why.

src/protocol/taiko_alethia/MessageRelayer.sol
```diff
@@ -0,0 +1,99 @@
...
+    {
+        address tipRecipient = TIP_RECIPIENT_SLOT.asAddress().tload();
+
+        uint256 valueToSend = msg.value - tip;
+        bool success;
+
+        if (gasLimit == 0) {
+            (success,) = to.call{value: valueToSend}(data);
+        } else {
+            require(gasLimit <= gasleft(), InsufficientGas());
```
ggonzalez94, line 92: I've been following the gas forwarding discussions. What about this simpler approach? I believe it should be good for all practical use cases(you have to send slightly more gas than absolutely needed, but not a problem for an honest relayer IMO)
```suggestion
            // EIP-150: Only 63/64 of gas can be forwarded to external calls.
            // Check against actual forwardable gas with buffer for operations before the call.
            uint256 maxForwardableGas = gasleft() * 63 / 64;
            require(gasLimit <= maxForwardableGas - 300, InsufficientGas());
```

src/protocol/taiko_alethia/MessageRelayer.sol
```diff
@@ -0,0 +1,99 @@
...
+            (success,) = to.call{value: valueToSend, gas: gasLimit}(data);
+        }
+
+        require(success, MessageForwardingFailed());
+        emit MessageForwarded(to, valueToSend, data);
+
+        TIP_RECIPIENT_SLOT.asAddress().tstore(address(0));
+
+        payable(tipRecipient).transfer(tip);
```
ggonzalez94, line 101: Why do we use `transfer` here? Isn't it more robust to do a `call` like in all our other instances? We are already making this function `nonReentrant`, so I don't see an issue

src/protocol/taiko_alethia/MessageRelayer.sol
```diff
@@ -0,0 +1,99 @@
...
+        IETHBridge.ETHDeposit memory ethDeposit,
+        uint256 height,
+        bytes memory proof,
+        address tipRecipient
+    ) external {
+        TIP_RECIPIENT_SLOT.asAddress().tstore(tipRecipient);
+
+        ethBridge.claimDeposit(ethDeposit, height, proof);
+
+        emit Relayed(ethDeposit, tipRecipient);
```
ggonzalez94, line 75: I believe we are emitting too many events unnecessarily. When a relay operation happens you have 3 events in total(1 for the bridge and 2 for this contract).
- We should of course keep the one on the bridge
- We can keep one of these 2. I would perhaps suggest the one on the `receiveMessage` function and maybe moved to the end of the call and we can include more parameters if necessary

ggonzalez94 (commented):
Looks great! I left a few comments and also added a diagram(which was helpful for me to remember the flow, I hope the same is true for other people).

One separate question I had is: should we allow the user to select the relayer? Right now the tip goes to whoever relays the message first on the destination chain. The problem with this is that it incentivizes a race condition, where relayers will spend gas to be the first(which might in turn result in worst pricing for users).
I imagine frontends might quote a tip for the user, and when the user accepts they could select who the designated relayer is.

src/protocol/taiko_alethia/MessageRelayer.sol
LeoPatOZ: okay updated it i removed the relayed event i think those interested can listen to the bridge to get the same info, then i added the message fowared event

src/protocol/taiko_alethia/MessageRelayer.sol
nikeshnazareth: Seems good to me, except I think it's more correct and easier to reason about if we:
- first estimate the `gasleft()` at the time of the call (accounting for the intermediate operations)
- ensure that amount is high enough to cover the EIP-150 discount

So it would be something like:

```solidity
uint256 maxForwardableGas = (gasleft() - BUFFER) * 63 / 64;
require(gasLimit <= maxForwardableGas, InsufficientGas());
```

PR #119, opened 2025-05-13
Signal service and bridge/state based tests

nikeshnazareth:
This PR updates the Signal Service and Bridge tests to be state based (similar to #85).

~~It is in an inconsistent state. I will update it tomorrow but I thought it would be useful to push the current version anyway.~~

The main advantages are:
- the test suite is easier to understand. There were a few times when I was helping review the functionality where I had to make one-off changes to the test suite without really tracking the overall structure. I think it's better encapsulated now.
- the test suite is more robust. All the off-chain work is isolated and it produces a config contract that can be imported into other tests. This is closer to how it would be used in practice (asynchronously). This means we do not need to handle foundry forks or connect to anvil. A lot of the offchain code can be removed, but the core idea of using anvil to construct storage proofs is retained. This also removes the unhelpful EVM-revert message that plagues the existing test suite.
- I think state-based tests are more flexible anyway. This makes it easier to capture different use cases, and to extend the test suite when new functionality is added.

I removed the previous signal and bridge tests. I believe we should also be able to remove the `deposit_signal_proof.rs` and `generic_signal_proof.rs` files, since they were used to get the constants that were copied to the previous tests. At the moment they just get a proof object (that has all the values that were previously printed) and then ignores it, but it could be easily updated to print those values if it is desirable to have a command line utility that prints out proofs for different signals / deposits. I'm not sure what that would be used for, but I decided not to remove the scripts just in case.

I think we can also remove the `start-anvil` and `stop-anvil` commands in the just file, but I'm not sure how to remove them completely (eg. should we get rid of anvil from `.github/workflows/test.yml`) so I just left them.


LeoPatOZ (approved):
I like the general approach the only nit i have is what we put in the universal tests, im not sure the ones currently are worth re-running in every test/state, although i agree with the concept of universal tests we should be more selective as to which ones to include because it does add extra load

PR #120, opened 2025-05-13
Add Cancel Eth Deposit on Bridge

LeoPatOZ:
## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a canceler address when creating deposits, allowing authorized users to cancel deposits if needed.
  - Introduced a new function to cancel deposits, enabling the designated canceler to claim the deposit on behalf of a recipient.
  - Added a new event to signal when a deposit is cancelled.

- **Tests**
  - Updated tests to cover the new canceler functionality and ensure correct handling of cancellable and non-cancellable deposits.


nikeshnazareth:
I'll review this properly today, but I have two weak objections to this idea (I could probably be convinced if there's a good reason):

- this seems like it's inconsistent with the ethos of making minimal solid components. Perhaps we should build a user-friendly bridge (possibly integrating it with MessageRelayer from #116), but it does seem to me like an additional optional layer that should be built on top of the core protocol.
- We can't just allow the deposit `from` address to cancel a deposit without address aliasing. And I'm not sure it's desirable to be opinionated about an aliasing mechanism (or even possible to do it safely with EIP-7702). It also makes the whole thing more convoluted because you would have to cancel the deposit from the source chain and then relay the cancellation on the destination.

On the second point, the basic problem is that address `X` on chain A may not be owned/controlled by the same entity that owns/controls address `X` on chain B. Consider a multisig wallet that has different owners, or a smart contract deployed on two chains where ownership was transferred on one chain (or it's a DAO that only responds to token holders on its own chain or whatever).

People sometimes try to identify EOA addresses by checking the codesize (the idea being that EOAs are controlled by off-chain private keys, so anyone who can access an EOA on one chain must already have access to that EOA on all other chains). If we go down this path we will have to be very careful and clear-headed about the possibilities for shenanigans. I suspect it would involve narrowing the number of addresses that can use this feature

nikeshnazareth:
I started reviewing this and then I realised you change it to Draft. I thought it makes sense to push my changes anyway for you consideration, but feel free to ignore/revert them as appropriate.

I also left the failing deposit tests because the correct way to fix them depends on whether #119 is merged

LeoPatOZ:
> I'm not sure what the `relayer` field is achieving. Should `claimDeposit` have access control so it can only be called by the relayer?

This was added in the relayer PR, the idea was to be able for the user to specify a specific relayer, if they wanted to or 0 for any

> * remove the `counterpart != address(0)` check?

I believe this was a merge bug


> * change `assertEq(X, false)` to `assert(X == true);`
>   
>   * the first syntax looks cleaner to me. Also, this is now doing the opposite check and I think that's why the test is failing.

I think this is still using the old test setup so none of these tests are working until we merge your state testing pr

src/protocol/ETHBridge.sol
```diff
@@ -62,15 +64,35 @@ contract ETHBridge is IETHBridge, ReentrancyGuardTransient {
 
     /// @inheritdoc IETHBridge
     function claimDeposit(ETHDeposit memory ethDeposit, uint256 height, bytes memory proof) external nonReentrant {
-        bytes32 id = _generateId(ethDeposit);
+        bytes32 id = _claimDeposit(ethDeposit, ethDeposit.to, ethDeposit.data, height, proof);
+        emit DepositClaimed(id, ethDeposit);
+    }
+
+    /// @inheritdoc IETHBridge
+    function cancelDeposit(ETHDeposit memory ethDeposit, address claimee, uint256 height, bytes memory proof)
```
pepebndc, line 72: Missing nonReentrant, might be worth adding to the `__claimDeposit` function to have it both in `cancelDeposit` and `claimDeposit`

PR #122, opened 2025-05-26
Merge inbox and pub feed

### LINKED ISSUE #93: Consider unifying the publication feed and the inbox
When we started working on the design, a shared publication feed where multiple rollups publish their blobs seemed like a powerful primitive for things like composability.
But since other rollups still need to interpret whatever is in that publication, and we don't want to require rollup proposers to have to run nodes for multiple rollups, most of the ideas we've had rely on higher level constructs(e.g. commitments).

If we end up not needing a separate publication feed, then it makes sense to unify them under a single contract for simplicity and efficiency

ggonzalez94:
## Description
Fixes #93 

This PR refactors the protocol to consolidate publication functionality by removing the separate `IPublicationFeed` interface and moving its functionality directly into `IInbox`. Since now we can assume all publications are related to a single rollup, this has some cascading benefits that we can use to simplify the `BaseProverMananger` and `CheckPointTracker` as well.

## List of Changes

- **Removed `IPublicationFeed` interface** - Consolidated publication feed functionality into `IInbox`
- **Refactored `BaseProverManager`** - Removed `publicationFeed` dependency, now uses `inbox` directly for publication validation
- **Simplified proof verification** - Removed `numPublications` parameter from various functions, since now it can be calculated internally
- In a previous PR we removed `start` from the `CheckPointTracker` and just used the latest proven checkpoint as the start. But this introduced an issue where you can get paid for already proven publications. Separately, we've always had the small issue that provers may get front run when generating a larger proof by someone posting a smaller proof that only covers a subset of publications.
To fix both of these, we now accept any (valid) start checkpoint, but only pay the prover from the latest proven checkpoint until the new checkpoint they are providing. 

## Open Questions

- Should we rename `TaikoInbox` into just a regular `Inbox` contract? I don't think it has too many Taiko specific things and could be used by other rollups
- There might still be an issue with `numDelayedPublications`, since we now allow a bigger range for the proof. This means that a prover can provide valid older delayed publications and still get paid more than they should?


src/protocol/BaseProverManager.sol
```diff
@@ -179,25 +175,32 @@ abstract contract BaseProverManager is IProposerFees, IProverManager {
...
-        checkpointTracker.proveTransition(start, end, numPublications, numDelayedPublications, proof);
+        ICheckpointTracker.Checkpoint memory lastProven = checkpointTracker.getProvenCheckpoint();
+        // Only count publications that have not been proven yet for the fee calculation
+        uint256 numPublications = lastPub.id - lastProven.publicationId;
+        require(
+            numDelayedPublications <= numPublications,
+            "Number of delayed publications cannot be greater than the total number of publications"
+        );
+
+        checkpointTracker.proveTransition(start, end, numDelayedPublications, proof);
```
LeoPatOZ, line 203: I think we can avoid reading the latestProven checkpoint here as we read it in the checkpoint tracker. Perhaps this check can be done in the checkpoint tracker:

``` solidity
uint256 numPublications = end.publicationId - latestProvenCheckpoint.publicationId;
require(numDelayedPublications <= numPublications, "Number of delayed publications cannot be greater than the total number of publications");

return numPublications
```
And we simply return the numPublications
WDYT?

nikeshnazareth (approved):

ggonzalez94:
> I'd also just like to confirm my understanding of the design

Yes, your description is accurate. I'll elaborate a bit on some of these points.

> the reason is so that a proof that advances the checkpoint by X publications is not invalidated by another proof that advances it by a smaller number of publications.

Correct, this should avoid the waste of a prover generating a larger proof and then it becoming invalid.

> during normal operation, we expect there is only one user that has any incentive to advance the checkpoint

Most of the times yes, but I see two situations where this might not be always true.
1. When proving a period where the original prover didn't prove on time and now all the provers are racing. A prover can use the strategy or proving a smaller subset to invalidate other provers work while he generates the rest of the proof.
2. When a prover is close to its deadline and has to generate a large proof. An attacker could submit a smaller proof just before theirs, invalidating their proof and causing them to miss their deadline.

> Separately, as I understand it, TaikoInbox.publish is no longer payable. It wasn't using msg.value and payPublicationFee is not payable. We assume any proposer would have prepaid enough to cover their publication fees. Is that right?

Correct, that was already the assumption but we had not made the changes. The proposer is expected to already have a balance. I don't remember exactly why we simplified this, but I remember it was intentional.

> I think this should have been built on https://github.com/OpenZeppelin/minimal-rollup/pull/115. I tried merging with that branch and rebasing on top of it, but neither option works cleanly. We'll need some due diligence to make sure resolving the merge conflicts does not introduce bugs.

One possible reason for the conflicts was that the branch you are mentioning already had a bunch of conflicts with main. I just solved those and merged to main. I'll do the same for this one and hopefully won't be too painful.

PR #124, opened 2025-06-09
Add some relayer tests

LeoPatOZ:
- Added some basic relayer tests as a sanity check

## Summary by CodeRabbit

- **New Features**
  - Added new test contracts and scenarios to improve coverage of relayer and deposit claim functionality, including relayed deposits to externally owned accounts and deposits claimable by relayers.
  - Introduced new state variables and contracts to facilitate relayer-related testing.
  - Expanded deposit specifications to include relayer test cases with structured calldata encoding.

- **Bug Fixes**
  - The `receiveMessage` function now explicitly checks that the provided Ether covers the tip amount, reverting with an error if insufficient.

- **Improvements**
  - Updated dependency for improved compatibility and performance.
  - Added a new error to the message relayer interface for clearer error reporting when the tip amount exceeds the provided value.
  - Enhanced Solidity interface definitions and ABI encoding for better interoperability.

- **Tests**
  - Added and updated test data, proofs, and scenarios to ensure robust validation of deposit and relayer logic.

### DIFF
--- a/offchain/deposit_signal_proof.rs
+++ b/offchain/deposit_signal_proof.rs
@@ -1,4 +1,4 @@
-use alloy::primitives::{address, bytes, U256};
+use alloy::primitives::{address, bytes, Address, U256};
 use eyre::Result;
 
 mod utils;
@@ -27,6 +27,8 @@ async fn main() -> Result<()> {
 
     let signal_service = deploy_signal_service(&provider).await?;
 
+    let zero_canceler = Address::ZERO;
+
     println!(
         "Deployed signal service at address: {}",
         signal_service.address()
@@ -37,7 +39,9 @@ async fn main() -> Result<()> {
     println!("Deployed ETH bridge at address: {}", eth_bridge.address());
 
     println!("Sending ETH deposit...");
-    let builder = eth_bridge.deposit(sender, data, bytes!()).value(amount);
+    let builder = eth_bridge
+        .deposit(sender, data, bytes!(), zero_canceler)
+        .value(amount);
     let tx = builder.send().await?.get_receipt().await?;
 
     // Get deposit ID from the transaction receipt logs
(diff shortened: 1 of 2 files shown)


nikeshnazareth:
My intuition is that these tests are testing the wrong aspect and should use a different framework. I can create a draft if you agree. One trivial point is that now that #139 exists, we should use that format. More importantly:

- I don't think the `MessageRelayer` tests should be integrated with the `ETHBridge` tests. The `ETHBridge` is a core part of the protocol while the `MessageRelayer` is not, so anyone should be able to ignore the relayer and still use and test the bridge. As it stands
  - all ETH bridge tests will now involve `messageRelayer` and `tipRecipient` values
  - the "relayer tests" in this PR are mostly about how the bridge behaves, not about how the relayer behaves.
- Much of the complexity of the ETHBridge tests are to construct a message on one chain, and then ensure the message is consumed correctly on a different chain. The relayer logic is all happening on the same chain.

When testing the `MessageRelayer`, I think we should assume the ETHBridge works correctly, start from the state where the messages are already claimable, and then do things like:
- ensure `relayMessage` triggers `receiveMessage`
- ensure the tip recipient set in `relayMessage` is paid in `receiveMessage` (unless the message itself contains a recipient)
- ensure the `claimDeposit` call can be used directly to trigger `recieveMessage` (once again, with both possibilities for zero/non-zero tip recipients).
- ensure the eventual recipient is invoked with the right data
- try all combinations of 
   - zero/non-zero tip recipient
   - zero/non-zero tip amount
   - empty/non-empty `data`
   - zero/non-zero gas limit
   - sufficient/insufficient funds sent with the message
   - sufficient/insufficient gas sent with the message
   - eventual call succeeds/fails
   - tip transfer succeeds/fails
   - the `TIP_RECIPIENT_SLOT` starts empty or with a non-zero value
- we should also make sure that the recipient can trigger `relayMessage` but
   - not reentrantly trigger `receiveMessage`
   - not change the value that the outer tip recipient gets, even though the `TIP_RECIPIENT_SLOT` is updated before the tip is transferred.

Issue #138, opened 2025-07-14
Replace (optional) `allowedRelayer` with `tipRecipient` in relayed message


nikeshnazareth:
Our current `MessageRelayer` contract does not perfectly enforce the `allowedRelayer` field.

Consider the scenario:
1. someone creates a messages with a particular `allowedRelayer`. This means they are expecting the `allowedRelayer` to call `relayMessage` and specify some tip recipient
2. an attacker calls `MessageRelayer.relayMessage` with some unrelated message (not destined for the `MessageRelayer`). This is still delivered as normal, since the bridge does not care who calls `claimDeposit`. However, the attacker can now set the tip recipient in the temporary storage of the message relayer.
3. within the same transaction, the attacker calls `ETHBridge.claimDeposit` on the target. This directly invokes `MessageRelayer.receiveMessage` (bypassing the `relayMessage` function and the `allowedRelayer` check) and sends the tip to the recipient from step 2.

Instead of attempting to fix the `allowedRelayer` mechanism, I think the user should actually just optionally specify a tip recipient. As I understand it, the original reason for the mechanism is that the user may create an out-of-band deal with a specific relaying entity, so we should allow that entity to perform the relay and collect the tip. However, specifying the address that can receive the tip should achieve the same purpose (and is simpler to enforce). It will still be possible for an arbitrary address to relay the message (and pay the gas costs), but the specified recipient will get the tip, so that does not seem like an attack.

With that in mind, my suggested change is:
- there is no `context` parameter. Anyone can call `relayMessage` and set the tip recipient in temporary storage
- the `receiveMessage` function accepts an optional `tipRecipient` parameter. If it is zero, then use the tip recipient from temporary storage.

In this way, the expected use cases are:
- if no tip recipient is set, use the current mechanism: a relayer calls `relayMessage` to set the recipient that will receive the tip in `receiveMessage`.
- if a tip recipient is set, the corresponding relayer (which could be a different address) calls `ETHBridge.claimDeposit`, which will trigger `receiveMessage` and distribute the tip accordingly.

The unexpected use cases are:
- anyone can call `ETHBridge.claimDeposit` on a message without a tip recipient. This will call `receiveMessage` and send the tip to whichever address happens to be saved in temporary storage. There is no incentive to do this, but there is no harm either (the message is delivered and there was no specified tip recipient who was bypassed).
[...]

PR #139, opened 2025-07-14
Fix Message Relayer

LeoPatOZ:
This fixes an issue described in https://github.com/OpenZeppelin/minimal-rollup/issues/138

## Summary by CodeRabbit

* **New Features**
  * Users can now specify a tip recipient address when relaying messages, allowing for more flexible tip distribution.

* **Refactor**
  * The tip recipient is now explicitly provided or retrieved from temporary storage, replacing previous implicit handling.
  * Relayer authorization checks based on the context field have been removed, enabling open message relaying.
  * Documentation and comments updated to clarify tip recipient handling and relayer permissions.

### DIFF
--- a/src/protocol/taiko_alethia/MessageRelayer.sol
+++ b/src/protocol/taiko_alethia/MessageRelayer.sol
@@ -6,7 +6,6 @@ import {TransientSlot} from "@openzeppelin/contracts/utils/TransientSlot.sol";
 import {IETHBridge} from "src/protocol/IETHBridge.sol";
 
 import {IMessageRelayer} from "src/protocol/IMessageRelayer.sol";
-import {MessageRelayer} from "src/protocol/taiko_alethia/MessageRelayer.sol";
 
 /// @dev Simple implementation of a message relayer.
 ///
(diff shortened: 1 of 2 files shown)


src/protocol/taiko_alethia/MessageRelayer.sol
```diff
@@ -21,32 +21,35 @@ import {MessageRelayer} from "src/protocol/taiko_alethia/MessageRelayer.sol";
...
+///                address(tipRecipient) // specified tip recipient
 ///                0.1 ether,        // tip for the relayer
 ///                0,                // gas limit
 ///                ""                // data (in this case empty)
 ///            )
 ///        )
 ///
 /// To relay the message:
-///    1. The allowed relayer calls `relayMessage` if set. Otherwise, anyone can call `relayMessage`.
+//     1. Anyone is allowed to call relayMessage however the tip recipient is determined the two following cases:
```
ggonzalez94, line 40:
```suggestion
//     1. Anyone is allowed to call `relayMessage` however the tip recipient is determined by one of the two following cases:
```

ggonzalez94: I think the natspec comments are lacking the third `/`

src/protocol/taiko_alethia/MessageRelayer.sol
```diff
@@ -21,32 +21,35 @@ import {MessageRelayer} from "src/protocol/taiko_alethia/MessageRelayer.sol";
...
+//     b) If a tip recipient is specified in the ETHDeposit message it will be used
+// It is up to the relayer to decide whether it is worth to relay this message or not (decided if they control the
+// tipRecipient address or not)
 ///    2. This will call claimDeposit on the ETHBridge
 ///    3. If the original message was specified correctly, this will call receiveMessage on this contract
-///    4. This will call the message recipient and send the tip to the relayer
+///    4. This will call the message recipient and send the tip to the tip recipient
 ///
-/// The relayer will net any tip minus the gas spent on the call to relayMessage.
+/// The tip recipient will net any tip minus the gas spent on the call to relayMessage.
```
ggonzalez94, line 49: not sure if this is 100% accurate anymore, since the recipient may actually be a different address. Perhaps we should get rid of this comment?

src/protocol/taiko_alethia/MessageRelayer.sol
```diff
@@ -62,31 +65,29 @@ contract MessageRelayer is ReentrancyGuardTransient, IMessageRelayer {
...
-    function receiveMessage(address to, uint256 tip, uint256 gasLimit, bytes memory data)
+    function receiveMessage(address to, uint256 tip, address tipRecipient, uint256 gasLimit, bytes memory data)
         external
         payable
         nonReentrant
     {
-        address tipRecipient = TIP_RECIPIENT_SLOT.asAddress().tload();
+        // If none specified use the one in temporary storage
+        if (tipRecipient == address(0)) {
+            tipRecipient = TIP_RECIPIENT_SLOT.asAddress().tload();
```
ggonzalez94, line 89: I think we should avoid overwriting the input variable for clarity/best practice

src/protocol/taiko_alethia/MessageRelayer.sol
```diff
@@ -62,31 +65,31 @@ contract MessageRelayer is ReentrancyGuardTransient, IMessageRelayer {
...
+        }
```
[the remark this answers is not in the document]
ggonzalez94: the tx will fail, so I don't think this is actually necessary

ggonzalez94 (commented):
Left a few minor comments, otherwise looks good to me

PR #145, opened 2025-07-24
removed _initializedTokens in favour of signal

pepebndc:
This PR is intended to verify the potencial removal of _initializedTokens to use the signal stored directly


LeoPatOZ:
Maybe this would be a terrible DX but to improve gas efficiency you could:
1) Just accept the signal / ID to check instead of token address --> refetching metadata --> calculate id. The user could generate this ID offchain, probably the most gas efficient but user unfriendly  
2) Allow user to pass in TokenInitialisation struct directly. This is slightly better in terms of DX and gas.

Issue #90, opened 2025-03-29
Support ERC20 and ERC721 Bridging


ggonzalez94:
While not core to the protocol, and more of an "application" concern, we agree that providing a bridge implementation for common token standards is important both for Taiko Alethia and for other rollups that want to use the stack.
Users of the stack can define if they want to use these implementations as the "canonical" bridge, or deploy their own.
- The implementations should use the SignalService for sending and verifying the messages
- We should support ETH + ERC token transfer(which shouldn't be that hard since Eth Bridging will be embedded into the SS) as the current implementation does.

Documentation for the current bridge implementation: https://docs.taiko.xyz/taiko-alethia-protocol/protocol-architecture/bridging

ggonzalez94:
Some open questions:
- Should it differentiate between tokens native to the L1 and native to the L2? e.g. the OP stack on L1 if the token is native to L1 locks it, and if it is native to L2 burns it
- What type of tokens should it support? e.g. the OP stack does not support tokens with transfer fees, rebasing tokens, and tokens with blocklists.

PR #146, opened 2025-07-25
ERC20, ERC721 and ERC1155 Bridge contracts and tests

pepebndc:
This PR introduces bridging support for ERC20, ERC1155, and ERC721 tokens, extending the existing ETH bridge functionality in the minimal rollup protocol.

The new bridges follow the same structure, flows, and systems as the ETHBridge contract, ensuring consistency and ease of maintenance. This includes deposit creation, claiming via storage proofs, and cancellation mechanisms, all integrated with the SignalService for cross-chain verification.

Closes issue https://github.com/OpenZeppelin/minimal-rollup/issues/90

### DIFF
--- a/src/protocol/ERC20Bridge.sol
+++ b/src/protocol/ERC20Bridge.sol
@@ -1,0 +1,247 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.28;
+
+import {BridgedERC20} from "./BridgedERC20.sol";
+import {IERC20Bridge} from "./IERC20Bridge.sol";
+import {ISignalService} from "./ISignalService.sol";
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {IERC20Metadata} from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {ReentrancyGuardTransient} from "@openzeppelin/contracts/utils/ReentrancyGuardTransient.sol";
+
+/// @title ERC20Bridge
+/// @notice A decentralized bridge for ERC20 tokens that allows anyone to initialize tokens
+/// @dev Uses a permissionless token initialization flow
+contract ERC20Bridge is IERC20Bridge, ReentrancyGuardTransient {
+    using SafeERC20 for IERC20;
+
+    /// @dev Signal type constants to differentiate signal categories
+    bytes32 private constant INITIALIZATION_SIGNAL_PREFIX = keccak256("ERC20_TOKEN_INITIALIZATION");
+    bytes32 private constant DEPOSIT_SIGNAL_PREFIX = keccak256("ERC20_DEPOSIT");
+
+    mapping(bytes32 id => bool processed) private _processed;
+    mapping(address token => bool initialized) private _initializedTokens;
+    mapping(bytes32 key => address deployedToken) private _deployedTokens;
+    mapping(address token => bool isBridgedToken) private _isBridgedTokens;
+
+    /// Incremental nonce to generate unique deposit IDs.
+    uint256 private _globalDepositNonce;
+
+    ISignalService public immutable signalService;
+
+    /// @dev Trusted source of commitments in the `CommitmentStore` that the bridge will use to validate withdrawals
+    /// @dev This is the Anchor on L2 and the Checkpoint Tracker on the L1
+    address public immutable trustedCommitmentPublisher;
+
+    /// @dev The counterpart bridge contract on the other chain.
+    /// This is used to locate deposit signals inside the other chain's state root.
+    /// WARN: This address has no significance (and may be untrustworthy) on this chain.
+    address public immutable counterpart;
+
+    constructor(address _signalService, address _trustedCommitmentPublisher, address _counterpart) {
+        require(_signalService != address(0), "Empty signal service");
+        require(_trustedCommitmentPublisher != address(0), "Empty trusted publisher");
+        require(_counterpart != address(0), "Empty counterpart");
+
+        signalService = ISignalService(_signalService);
+        trustedCommitmentPublisher = _trustedCommitmentPublisher;
+        counterpart = _counterpart;
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function processed(bytes32 id) public view returns (bool) {
+        return _processed[id];
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function isTokenInitialized(address token) public view returns (bool) {
+        return _initializedTokens[token];
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function isInitializationProven(bytes32 id) public view returns (bool) {
+        return _processed[id];
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function getDeployedToken(address originalToken) public view returns (address) {
+        bytes32 key = keccak256(abi.encode(originalToken));
+        return _deployedTokens[key];
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function getInitializationId(TokenInitialization memory tokenInit) public pure returns (bytes32 id) {
+        return _generateInitializationId(tokenInit);
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function getDepositId(ERC20Deposit memory erc20Deposit) public pure returns (bytes32 id) {
+        return _generateDepositId(erc20Deposit);
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function initializeToken(address token) external returns (bytes32 id) {
+        require(token != address(0), "Invalid token address");
+        require(!_initializedTokens[token], "Token already initialized");
+
+        // Read token metadata with fallbacks for non-compliant tokens
+        string memory name;
+        string memory symbol;
+        uint8 decimals;
+
+        try IERC20Metadata(token).name() returns (string memory _name) {
+            name = _name;
+        } catch {
+            name = "Unknown Token Name";
+        }
+
+        try IERC20Metadata(token).symbol() returns (string memory _symbol) {
+            symbol = _symbol;
+        } catch {
+            symbol = "UNKNOWN";
+        }
+
+        try IERC20Metadata(token).decimals() returns (uint8 _decimals) {
+            decimals = _decimals;
+        } catch {
+            decimals = 18; // Standard default
+        }
+
+        TokenInitialization memory tokenInit =
+            TokenInitialization({originalToken: token, name: name, symbol: symbol, decimals: decimals});
+
+        id = _generateInitializationId(tokenInit);
+
+        // Mark token as initialized locally
+        _initializedTokens[token] = true;
+
+        // Send signal for cross-chain initialization
+        signalService.sendSignal(id);
+
+        emit TokenInitialized(id, tokenInit);
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function deployCounterpartToken(TokenInitialization memory tokenInit, uint256 height, bytes memory proof)
+        external
+        returns (address deployedToken)
+    {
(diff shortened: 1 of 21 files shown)


src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,247 @@
...
+
+    /// @inheritdoc IERC20Bridge
+    function getDepositId(ERC20Deposit memory erc20Deposit) public pure returns (bytes32 id) {
+        return _generateDepositId(erc20Deposit);
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function initializeToken(address token) external returns (bytes32 id) {
+        require(token != address(0), "Invalid token address");
+        require(!_initializedTokens[token], "Token already initialized");
```
LeoPatOZ, line 86: worth mentioning in this PR since other comments were lost, i am in favour of removing this mapping but keep the  `isTokenInitialized` view function for dapps to use.

src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,247 @@
...
+            // No bridged token found → transfer the original token we're holding
+            IERC20(erc20Deposit.localToken).safeTransfer(to, erc20Deposit.amount);
+        }
+    }
+
+    /// @dev Checks if a token is a bridged token deployed by this bridge.
+    /// @param token The token address to check
+    /// @return true if the token is a bridged token deployed by this bridge
+    function _isBridgedToken(address token) internal view returns (bool) {
+        return _isBridgedTokens[token];
```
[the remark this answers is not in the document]
LeoPatOZ: but tokens can only ever be marked as true in the `_isBridgedToken` mapping if you called the `deployCounterpartToken` function - in that case it would also be in the `_deployedTokens`? This means the `_deployedTokens` mapping serves the dual purpose 
`_isBridgedToken(0x456)` would return true because it was deployed in the `deployCounterpartToken` flow and therefore would have a non zero address associated with it

src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function deposit(address to, address localToken, uint256 amount) external nonReentrant returns (bytes32 id) {
+        // Check if token is initialized (for original tokens) or is a bridged token deployed by this bridge
+        require(_initializedTokens[localToken] || _isBridgedToken(localToken), TokenNotInitialized());
+
+        // If depositing an original token, use its address directly
+        address originalToken = localToken;
+        if (_isBridgedToken(localToken)) {
```
LeoPatOZ, line 160: we check _isBridgedToken twice, maybe we can just save it

bool `isBridgedToken` = _isBridgedToken(localToken)`

pepebndc: No longer aplicable after the changes on https://github.com/OpenZeppelin/minimal-rollup/pull/146/commits/8ecd8f45e7d25718550ad59f74d69fd66284d999

src/protocol/IERC20Bridge.sol
```diff
@@ -0,0 +1,116 @@
...
+
+    struct ERC20Deposit {
+        // The nonce of the deposit
+        uint256 nonce;
+        // The sender of the deposit
+        address from;
+        // The receiver of the deposit
+        address to;
+        // The original ERC20 token address (always refers to the original token, not bridged)
+        address localToken;
```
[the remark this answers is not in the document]
pepebndc: done https://github.com/OpenZeppelin/minimal-rollup/pull/146/commits/f843d25d24acd05afc83e1ffa8d9ef5c1d3343b9

src/protocol/IERC20Bridge.sol
```diff
@@ -0,0 +1,109 @@
...
+    /// @param token The ERC20 token address to initialize
+    function initializeToken(address token) external returns (bytes32 id);
+
+    /// @dev Proves a token initialization from the source chain and deploys the bridged token.
+    /// @param tokenInit The token initialization data
+    /// @param height The height of the checkpoint on the source chain
+    /// @param proof Encoded proof of the initialization signal
+    function deployCounterpartToken(TokenInitialization memory tokenInit, uint256 height, bytes memory proof)
+        external
+        returns (address deployedToken);
```
nikeshnazareth, line 95: I think we should update all functions/structs/events to consistently describe these operations as (1) logging the token description and (2) deploying the counterpart token. For instance:

- I would change `TokenInitialization` and `TokenInitialized` to `TokenDescription` and `TokenDescriptionRecorded` or similar
- I don't think we need an `isInitializationProven` function if it's just checking that a particular `id` is `processed`. Anyone who knows what that ID is will know that it applies to the token description
- `TokenInitializationProven` should be `CounterpartTokenDeployed`  (or `BridgedTokenDeployed`) since that is the actual action taken by the function. Similarly, the `InitializationAlreadyProven` error should be `CounterpartTokenAlreadyDeployed`

src/protocol/IERC20Bridge.sol
```diff
@@ -0,0 +1,109 @@
...
+    /// @param deposit The ERC20 deposit
+    event DepositMade(bytes32 indexed id, ERC20Deposit deposit);
+
+    /// @dev Emitted when a deposit is claimed.
+    /// @param id The deposit id
+    /// @param deposit The claimed ERC20 deposit
+    event DepositClaimed(bytes32 indexed id, ERC20Deposit deposit);
+
+    /// @dev Failed to claim the deposit.
+    error FailedClaim();
```
nikeshnazareth, line 57: I think this is an unused error

pepebndc: indeed, removed https://github.com/OpenZeppelin/minimal-rollup/pull/146/commits/cfaec5808bf26a4d841bc95503ecda8f17a94483

src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,244 @@
...
+        // Mark initialization as processed
+        _processed[id] = true;
+
+        // Deploy the bridged token
+        deployedToken =
+            address(new BridgedERC20(tokenInit.name, tokenInit.symbol, tokenInit.decimals, tokenInit.originalToken));
+
+        // Store the mapping
+        bytes32 key = keccak256(abi.encode(tokenInit.originalToken));
+        _deployedTokens[key] = deployedToken;
```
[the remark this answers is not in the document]
nikeshnazareth: I came to make this comment until I saw coderabbit had made it and you resolved it. Do you disagree?

Note that we cannot guarantee that the `_processed[id]` check will prevent overwriting this because it's possible for someone to deploy a token and then later upgrade it to have a different name/symbol/decimals. We could potentially treat that as a new token (so we would have different bridged tokens for each configuration of the original token) but that would add some complexity about how to handle the `_deployedTokens` mapping. It's probably easier to just prevent this by refusing to deploy a token if the `originalToken` address has already been bridged.

src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,233 @@
...
+        // Handle token transfer based on whether it's a bridged token or original token
+        IERC20(localToken).safeTransferFrom(msg.sender, address(this), amount);
+        if (_isBridgedToken(localToken)) {
+            // This is a bridged token being sent back to its origin, burn it
+            BridgedERC20(localToken).burn(amount);
+        }
+
+        // Send signal
+        signalService.sendSignal(id);
+        emit DepositMade(id, erc20Deposit);
```
nikeshnazareth, line 179: this event should probably mention the `localToken` as well if it's different from the original

src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function claimDeposit(ERC20Deposit memory erc20Deposit, uint256 height, bytes memory proof) external nonReentrant {
+        bytes32 id = _claimDeposit(erc20Deposit, erc20Deposit.to, height, proof);
```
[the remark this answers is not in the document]
pepebndc: changed it in the erc20 bridge, but kept in 721 and 1155 due to them having a cancel flow and the internal function being useful

https://github.com/OpenZeppelin/minimal-rollup/pull/146/commits/e3eda09c6cd1e6ce5dd78e1eda8b34efb3ec18ae

src/protocol/IERC20Bridge.sol
pepebndc: https://github.com/OpenZeppelin/minimal-rollup/pull/146/commits/6361902e5545fb5d6064c37621bb64586c240fe2

LeoPatOZ:
Nit: change the test names to camel_snakeCase

src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        signalService.sendSignal(id);
+        emit DepositMade(id, erc20Deposit, localToken);
+    }
+
+    /// @inheritdoc IERC20Bridge
+    function claimDeposit(ERC20Deposit memory erc20Deposit, uint256 height, bytes memory proof) external nonReentrant {
+        bytes32 id = _generateDepositId(erc20Deposit);
+        require(!processed(id), AlreadyClaimed());
```
LeoPatOZ, line 160:
```suggestion
        require(!_processed[id], AlreadyClaimed());
```
similar to  L113

src/protocol/ERC20Bridge.sol
```diff
@@ -0,0 +1,200 @@
...
+
+        _processed[id] = true;
+        _sendERC20(erc20Deposit);
+
+        emit DepositClaimed(id, erc20Deposit);
+    }
+
+    /// @dev Function to transfer ERC20 to the receiver.
+    /// @param erc20Deposit The deposit information containing the original token address
+    function _sendERC20(ERC20Deposit memory erc20Deposit) internal {
```
LeoPatOZ, line 172: should we just remove this function and just dd it to the claimDeposit ?

nikeshnazareth:
I made some changes to the bridge. I suspect the 1155 bridge could be simplified further by:
- removing the `data` parameter in `mint`
- removing the special case for empty URIs (compare the `_sendERC1155` function to the `_sendERC721`) function

However, I am not familiar with the 1155 spec so I'm not quite sure how to handle the URIs (in particular, it seems there is a contract-wide URI, which might need to be taken into account) and I didn't want to start messing with a system that I don't understand. I also haven't reviewed the tests, but I suspect we can reuse the same structure that we have for the existing ETH bridge to simplify the test suite and expand its comprehensiveness. I'll probably chase both of these down at some point but I don't want to be a blocker on this, so I will approve the PR now.

It also seems to me that if we're going to have a 1155 bridge then we should support batch transfers, because that seems to be one of the main use cases for the standard, but we can add that in a later PR.

src/protocol/ERC721Bridge.sol
```diff
@@ -0,0 +1,242 @@
...
+        returns (address deployedToken)
+    {
+        bytes32 id = _generateTokenDescriptionId(tokenDesc);
+        require(!_processed[id], CounterpartTokenAlreadyDeployed());
+
+        signalService.verifySignal(height, trustedCommitmentPublisher, counterpart, id, proof);
+
+        _processed[id] = true;
+
+        require(
```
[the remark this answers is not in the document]
nikeshnazareth: I believe this is in commit 4f2f2d3

src/protocol/ERC721Bridge.sol
```diff
@@ -0,0 +1,242 @@
...
+        try IERC721Metadata(originalToken).tokenURI(tokenId) returns (string memory uri) {
+            tokenURI = uri;
+        } catch {}
+
+        address actualOriginalToken;
+        if (isBridged) {
+            actualOriginalToken = BridgedERC721(originalToken).originalToken();
+        } else {
+            actualOriginalToken = originalToken;
+        }
```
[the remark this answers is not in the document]
nikeshnazareth: I believe this is in commit 4f2f2d3

Issue #126, opened 2025-06-09
Resolve circular dependencies when deploying


ggonzalez94:
There are a couple instances where we have circular dependencies, where contracts reference each other's address in their constructor, making it more challenging to deploy them.

Concretely these are the two cases:

- ProverManager ↔ CheckpointTracker

- ProverManager ↔ Inbox (TaikoInbox)

There's a few ways we can go around these, which I'll list from requiring less changes to the protocol to requiring more.

1. Write a fancy deployment script that uses CREATE2 cleverly to achieve this
2. Use initializer functions in some of the contracts to break out of the circular dependency loop.
3. Re architect them to remove the circular dependency

PR #149, opened 2025-07-29
Attempt to remove circular dependencies

LeoPatOZ:
This is an example of how to fix https://github.com/OpenZeppelin/minimal-rollup/issues/126. 

Main issue is adding only modifier which adds overhead. 

To make it even more 'trustless' could set owner to zero address after initialising prover manager. 

## Summary by CodeRabbit

* **New Features**
  * Added explicit, owner-controlled initialization for key contract addresses in CheckpointTracker and TaikoInbox, enhancing security and flexibility.
  * Introduced new events to signal when critical addresses are initialized or updated.
  * Added new functions allowing owners to update these addresses post-deployment.

* **Tests**
  * Expanded and updated tests to cover new initialization logic, access controls, and address updates.


src/libs/LibProvingPeriod.sol
```diff
@@ -52,7 +52,7 @@ library LibProvingPeriod {
         period.stake = stake;
     }
 
-    /// @dev Whether the period has been initialized
+    /// @dev Whether the period has been initialised
```
pepebndc, line 55: nit: if you write it with a `s` here, the function name should also be with an `s`

src/protocol/taiko_alethia/TaikoInbox.sol
```diff
@@ -39,21 +47,26 @@ contract TaikoInbox is IInbox, DelayedInclusionStore {
...
-        proposerFees = IProposerFees(_proposerFees);
 
         // guarantee there is always a previous hash
         _publicationHashes.push(0);
     }
 
     /// @inheritdoc IInbox
-    function publish(uint256 nBlobs, uint64 anchorBlockId) external {
+    function initializeProposerFees(address _proposerFees) external onlyOwner {
+        require(address(proposerFees) != address(0), "ProposerFees already initialised");
```
pepebndc, line 62:
```suggestion
        require(address(proposerFees) == address(0), "ProposerFees already initialised");
```

pepebndc (changes requested):

src/libs/LibProvingPeriod.sol
pepebndc: Applies to the rest of the instances changed in the PR

nikeshnazareth:
Sorry, for being slow on this - I just caught up. It seems a bit opinionated to me, so I have some suggestions:

Firstly, what was the rationale for removing `Ownable` (ef30453)? If we're going to set a privileged deployer address and create an `onlyDeployer` modifier, I would have thought `Ownable` would be the simplest/cleanest solution.

But more importantly, I don't think we need to insist on any prover or proposer fees initializer requirements. As I understand it:
- the Checkpoint Tracker is self-contained and functional whether or not there is a prover manager. The [access control line](https://github.com/OpenZeppelin/minimal-rollup/blob/ef30453396067c083f5dd7879436e29ae39e9898/src/protocol/CheckpointTracker.sol#L76) even allows there to be no prover manager.
- similarly the inbox is self-contained and functional, whether or not there is a prover manager. We could follow the same pattern of only calling `payPublicationFee` if the prover manager exists.
- the dependency is created when we add the prover manager, which ties those two contracts together economically (but not functionally). The prover manager is not self-contained, and is only functional if it references an inbox and a checkpoint tracker.

So my suggested pattern is:
- the `CheckpointTracker` and `TaikoInbox` should both be `Ownable`, where the owner has the ability to set the prover manager in both cases
- we do not insist that either contract must set the prover manager. We also do not insist that the prover manager cannot be changed once it's set. The owner can update the prover manager at will, and can renounce their role whenever they want to remove that power.

If anyone wants to deploy all contracts together atomically, they should create a single transaction that:
- deploys the tracker and inbox
- deploys the ProverManager, passing in the tracker and inbox addresses
- sets the prover manager in both tracker and inbox contracts
- renounces ownership of both contracts

What do you think? I'm happy to make the change if you agree.

LeoPatOZ:
@nikeshnazareth i made the changes you proposed please have a look!

nikeshnazareth (approved):

